PR #48, opened 2021-12-24
Update translation_nl.xml

Thodor12:
Edited all of the Dutch translations.
Only "CP_vehicle_courseGeneratorSetting_centerMode_lands" was unclear in terms of meaning/usage, would like to have some more info on that one to provide a proper translation.


XiniX38:
> Edited all of the Dutch translations.
> Only "CP_vehicle_courseGeneratorSetting_centerMode_lands" was unclear in terms of meaning/usage, would like to have some more info on that one to provide a proper translation.

After viewing this YT video "Courseplay FS22: first version" https://youtu.be/_kyAaN2r8Uk (in which Tensuko shows us an excellent introduction  of the new integrated UI), I would suggest "Veldsecties" as a Dutch translation for "Lands" as the "Lands" option actually subdivides a field in smaller work sections of that field.

pvaiko (changes requested):
Please resolve conflicts

PR #52, opened 2021-12-24
Simplify Github workflow code

Thodor12:
Your workflow could use a bit of love :D

I also included the event for `pull_request` so that people who want to contribute using PR's also need to have the unit tests go correctly.

Additionally I made sure that builds only trigger when any lua file has changed (so you're not unnecessarily triggering unit tests for updating a README or something similar)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ on:
 # The list of jobs this workflow executes
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: windows-latest
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


Thodor12:
When merging you can use the squash commits feature to avoid having a bunch of individual commits be put on main as opposed to only getting 1 commit: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-pull-request-commits

pvaiko:
@Thodor12 thanks, I'm familiar with GitLab CI/CD only and did not have the motivation/time to clean this up properly. And yes, we know how squash works :)

PR #101, opened 2021-12-25
Update translation_fr.xml

Lactic68:
In order to close conflicts


Lactic68 (commented):
Layout change to mirror EN layout

Issue #28, opened 2021-12-24
Setting Fieldspeed etc.


punchi21:
I can't set the Fieldspeed, wont change
And reservespeed only goes between 2 and 5 km/h 

The thing is, this only accurs on an old savegame. On a fresh savegame (same mods) it works without problems

hanrahs:
I had the same issue but on a new game also, both the fieldwork speed and the on field speed are stuck on 5km/h, reverse and turning you could go between 2 and 5km/h. restarting the game made no difference.

Alaxsxag:
same here...

Horozor:
Same. But only with grass mower, combine works fine. 10km/h

DTrigger05:
I currently also have this problem, with the John Deere T560. Have not tested any other vehicles yet.

Tensuko:
What Version did you all use ?

We pushed something with a speed fix.
We need to figure our on what condition the speed gets limited.

DTrigger05:
Originally noted with 7.0.0.1. Problem persists in 7.0.0.3, does not appear to be dependent on a specific vehicle, tried with two others and still limited to 3mph in fieldwork/field speed settings. Running an existing pre-courseplay save.

Peiren85:
> What Version did you all use ?
> 
> We pushed something with a speed fix. We need to figure our on what condition the speed gets limited.

Ich  kann in dem Fall nur für mich sprechen aber ich habe alle Versionen ausprobiert. Also von der 7.0.0.1 bis hin zur 7.0.0.3 . Hab sogar versucht den Mod komplett zu entfernen und dann noch mal neu drauf zu machen ab er keine Besserung. Ging bei mir auch von Anfang an nicht.

Horozor:
7.0.0.1

punchi21:
I used version 7.0.0.1

Horozor:
7.0.0.3 same, 5km/h

hanrahs:
updating to 7.0.0.3 fixed it for me (in the save I was having issues with)

Opa-Andre:
The speed issue in 7.0.0.3 is obviously due to using the wrong variable... I just tested it. My turning speed was set to 15km/h but CP was using 5km/h only. As the "reverse speed" was set to 5km/h (but my tractor didn't need to drive backwards) I set this reverse speed setting to 10 km/h for testing - and from there on, my tractor was driving 10 km/h when changing lanes...

geati6783:
[...]

PR #119, opened 2021-12-26
Update AIParameterSpeedSetting.lua

Horozor:
Fix for speed settings #28


pvaiko (commented):
Thanks, this seems to fix the issue but is only a band aid. We'll push a fix soon.

PR #121, opened 2021-12-26
Update translation_ru.xml

Gonimy-Vetrom:


Gonimy-Vetrom:
Many small and not minor changes.

PR #215, opened 2021-12-28
Update german translation

StefanBraun2001:
Updated german translations, including a few fixes. The first two actions are just beginner problems and can be ignored


StefanBraun2001:
Closes, since i made a mistake.

PR #305, opened 2022-01-02
added field margin setting to course generation

sbabiy:
Adds a setting for field outer margin to minimize missed crops in some cases. In particular, with smooth and rounded course generation

### DIFF
--- a/scripts/courseGenerator/CourseGeneratorInterface.lua
+++ b/scripts/courseGenerator/CourseGeneratorInterface.lua
@@ -73,36 +73,6 @@ function CourseGeneratorInterface.generate(fieldPolygon,
 	local islandNodes = Island.findIslands(Polygon:new(CourseGenerator.pointsToXy(fieldPolygon)))
 
 	--------------------------------------------------------------------------------------------------------------------
-	-- Field Margin
-	-----------------------------------------------------------------------------------------------------------------------
-	poly = Polygon:new(CourseGenerator.pointsToXy(fieldPolygon))
-	local centerX, centerZ = poly:getCenter()
-	centerZ = centerZ * -1
-	
-	local fieldPolygonWithMargin = {}
-
-	if fieldMargin > 0 then
-		for i, point in ipairs(fieldPolygon) do
-			local p = shallowCopy(point)
-			if p.x > centerX then
-				p.x = p.x + fieldMargin
-			else
-				p.x = p.x - fieldMargin
-			end
-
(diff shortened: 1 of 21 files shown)


scripts/courseGenerator/CourseGeneratorInterface.lua
```diff
@@ -71,6 +72,36 @@ function CourseGeneratorInterface.generate(fieldPolygon,
...
+	poly = Polygon:new(CourseGenerator.pointsToXy(fieldPolygon))
+	local centerX, centerZ = poly:getCenter()
+	centerZ = centerZ * -1
+	
+	local fieldPolygonWithMargin = {}
+
+	if fieldMargin > 0 then
+		for i, point in ipairs(fieldPolygon) do
+			local p = shallowCopy(point)
+			if p.x > centerX then
```
pvaiko, line 87: This works fine for rectangular and probably most convex polygons only.

scripts/courseGenerator/CourseGeneratorSettings.lua
```diff
@@ -317,6 +317,22 @@ function HeadlandOverlapPercent:init(vehicle)
 	self:set(7)
 end
 
+--- Field Margin
+---@class FieldMargin
+FieldMargin = CpObject(SettingList)
```
pvaiko, line 322: This is not needed in FS22, the definition in XML is sufficient.

pvaiko (changes requested):
Thanks for the PR!

I think the easiest way to solve this would be using the existing calculateHeadlandTrack function to generate the new field boundary.  

Also, we should enable -2 to +5 values for the margin with maybe 0.2 steps, negative values increasing the size of the field, positive ones decreasing.

Let me know if you don't feel comfortable making these changes.

scripts/courseGenerator/CourseGeneratorInterface.lua
sbabiy: Thanks for your feedback! I reviewed the calculateHeadlandtrack function, but I'm not certain on the best way to incorporate this into it. Any help you could provide would be greatly appreciated!

pvaiko: No problem, I created a field-margin branch, please create a new pull request to that branch (instead of main) and we go from there.

sbabiy:
@pvaiko please take a look at the latest PR changes. It's working with convex polygons, but I'm not certain if that is the right approach.

PR #322, opened 2022-01-02
Update VehicleConfigurations.xml

RealNeo80:
should fix basic problems with titan18, so far without using headlands

### DIFF
--- a/config/VehicleConfigurations.xml
+++ b/config/VehicleConfigurations.xml
@@ -125,10 +125,12 @@ You can define the following custom settings:
              turnRadius = "9"
     />
     <!--\vehicles\lemken-->
-    <Vehicle name="titan11.xml"
+    <Vehicle name="titan18.xml"
+			 workingWidth = "4.2"
              noReverse="true"
-             turnRadius = "4.5"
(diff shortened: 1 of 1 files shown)


config/VehicleConfigurations.xml
```diff
@@ -125,10 +125,12 @@ You can define the following custom settings:
...
     />
     <!--\vehicles\lemken-->
-    <Vehicle name="titan11.xml"
+    <Vehicle name="titan18.xml"
+			 workingWidth = "4.2"
              noReverse="true"
-             turnRadius = "4.5"
+             turnRadius = "7.5"
              implementWheelAlwaysOnGround = "true"
+			 turnAnimTime="3.500000"
```
pvaiko, line 133: I'm not aware of such a config option

pvaiko (changes requested):

config/VehicleConfigurations.xml
RealNeo80: don´t ask me why it wasn´t waiting on the field edge to turn the plough without this option, but now its working without %)

PR #334, opened 2022-01-03
Update GlobalSettingsSetup.xml

Roby1164:
Failure to display the set data in the menu


pvaiko (approved):
Thanks for the PR, they do show up for me, and spaces are allowed by the standard, so should not make any difference, but it is certainly more readable.

PR #346, opened 2022-01-04
Feature field margin

sbabiy:
@pvaiko please take a look at the latest PR changes. It's working with convex polygons, but I'm not certain if that is the right approach.

### DIFF
--- a/translations/translation_en.xml
+++ b/translations/translation_en.xml
@@ -139,6 +139,10 @@
 
 		<text name="CP_vehicle_courseGeneratorSetting_headlandOverlapPercent_title"						text="Headland overlap" />
 		<text name="CP_vehicle_courseGeneratorSetting_headlandOverlapPercent_tooltip"					text="Overlap headland tracks as percent of working width" />
+
+		<text name="CP_vehicle_courseGeneratorSetting_fieldMargin_title"								text="Field margin" />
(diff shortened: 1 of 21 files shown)


translations/translation_en.xml
```diff
@@ -139,6 +139,10 @@
 
 		<text name="CP_vehicle_courseGeneratorSetting_headlandOverlapPercent_title"						text="Headland overlap" />
 		<text name="CP_vehicle_courseGeneratorSetting_headlandOverlapPercent_tooltip"					text="Overlap headland tracks as percent of working width" />
+
+		<text name="CP_vehicle_courseGeneratorSetting_fieldMargin_title"								text="Field margin" />
+		<text name="CP_vehicle_courseGeneratorSetting_fieldMargin_tooltip"								text="Enlarge field size by the margin to reduce unworked edges" />
```
Tensuko, line 144: "Enlarge or shrink field size to create an overlap to the fieldborder or to have some savedistance to objects that are to close to the fieldborder."

Something like that.

pvaiko:
Thanks, I refactored it a bit, see here: https://github.com/Courseplay/Courseplay_FS22/commit/02659a1dd09af43f2f0d30c8b8f44f85083f717a

sbabiy:
Thanks @pvaiko! I tried something similar but was running into crashes when margin exceeded certain negative or positive values, it would return nil for generated course. This is very helpful!!!

PR #429, opened 2022-01-08
Same UI as in FS 19 

R8shabt:
The UI in fs 19 was so perfect. If the UI could be like that 1 that would be great


Tensuko:
1st this is PR section, 2nd read wiki, it wont happen.

PR #484, opened 2022-01-11
WIP

schwiti6190:
- Improved the ai event handling for implements, like for example: "onAIImplementStart"
- Added missing implements to the aiImplementList, still WIP
- Added a few missing implement controllers


scripts/ai/controllers/ImplementController.lua
```diff
@@ -4,6 +4,18 @@ ImplementController = CpObject()
...
+end
+
+function ImplementController:delete()
+    self:unregisterOverwrittenFunctions()
+    self:unregisterAIEvents()
+end
+
+function ImplementController:setAllStaticParameters(settings,fillLevelManager)
+    self.settings = settings
+    self.fillLevelManager = fillLevelManager
```
pvaiko, line 18: We set this but then don't seem to be using for everything. We have to make a decision how we want to handle fill levels. We can move it to the controllers, but not sure how to handle if the fill type is not on the same implement, or on the vehicle directly etc (in other words, fill levels are vehicle global and not necessarily local to the specific implement, therefore something like the FillLevelManager may make sense.

scripts/ai/controllers/FertilizingSowingMachineController.lua
```diff
@@ -0,0 +1,32 @@
...
+	end
+	self:registerOverwrittenFunction(self.sowingMachine,"processSowingMachineArea",processSowingMachineArea)
+end
+
+function FertilizingSowingMachineController:update()
+	local maxSpeed
+	if self.settings.sowingMachineFertilizerEnabled:getValue() then 
+		local fillUnitIndex = self.sowingMachine:getSprayerFillUnitIndex()
+		if self.sowingMachine:getFillUnitFillLevel(fillUnitIndex) <= 0 then 
+			SpecializationUtil.raiseEvent(self.vehicle,"onCpEmptyOrFull")
```
pvaiko, line 25: Anything preventing us from making two different events, one for empty, one for full? Would probably be easier to react on them.

pvaiko (changes requested):

scripts/ai/controllers/ImplementController.lua
schwiti6190: I understand your idea behind the FillLevelManager. My idea was to seperate these for now, as we only need to know if we can continue the work for now and this is diffrent for each implement. For example the possible forage wagon needs to ignore the additional filltype (silage additives) or the sowing machine might need to ignore the fertilizer depending on the setting for it.

scripts/ai/AIDriveStrategyFindBales.lua
```diff
@@ -122,6 +122,7 @@ end
 --- Static parameters (won't change while driving)
 -----------------------------------------------------------------------------------------------------------------------
 function AIDriveStrategyFindBales:setAllStaticParameters()
+    AIDriveStrategyFindBales:superClass().setAllStaticParameters(self)
```
pvaiko, line 125: What about the initializeImplementControllers function, is that still needed?

PR #504, opened 2022-01-12
Hungarian Translastion

RealNeo80:
on behalf of Laczkovics111


Tensuko:
I just added something missing and got conflict now, I add it directly...

PR #541, opened 2022-01-15
Update translation_ru.xml

Gonimy-Vetrom:


Tensuko:
Thanks for translation, but can we pls stick to out formatation?
Corrected that now.

Gonimy-Vetrom:
Okay. Just because of a large number of spaces, my text does not fit on the screen. :(

Tensuko:
Our translation names might get longer depending on what we will implement, changing all translations after that gives us more work later then.

What text tool do you use ?
I use VS Code or Notepad++ it should automatic scroll to the right if you keep type.

Gonimy-Vetrom:
Notepad++

PR #550, opened 2022-01-16
Issue 451

schwiti6190:


scripts/ai/AIDriveStrategyFieldWorkCourse.lua
```diff
@@ -67,7 +67,9 @@ function AIDriveStrategyFieldWorkCourse:start(course, startIx)
     if distance > 2 * self.turningRadius then
         self:debug('Start waypoint is far (%.1f m), use an alignment course to get there.', distance)
         self.course = course
-        self:startAlignmentTurn(course, startIx, 1)
+        --- Find the closest waypoint and create a alignment course form there to the start ix.
+        local _, _, ixClosestRightDirection, _ = course:getNearestWaypoints(self.vehicle:getAIDirectionNode())
+        self:startAlignmentTurn(course, ixClosestRightDirection, startIx)
```
schwiti6190, line 72: @pvaiko Do you think this is a good idea?

scripts/ai/AIDriveStrategyCourse.lua
```diff
@@ -278,4 +278,13 @@ end
...
 function AIDriveStrategyCourse:isCloseToCourseStart(distance)
     return self.course:getDistanceFromFirstWaypoint(self.ppc:getCurrentWaypointIx()) < distance
-end
\ No newline at end of file
+end
+
+function AIDriveStrategyCourse:delete()
+    if self.course then 
+        --- Remembers the third last waypoint ix driven.
+        local ix = math.max(1,self.course:getCurrentWaypointIx()-2)
```
pvaiko, line 286: We have function for this: rememberWaypointToContinueFieldwork() as it isn't trivial

scripts/ai/AIDriveStrategyFieldWorkCourse.lua
pvaiko: Don't think is needed, `self:startAlignmentTurn(course, startIx, startIx)` should be fine, it'll start the alignment course at the vehicle's position anyway, see comment at startAlignmentTurn

pvaiko (changes requested):

scripts/ai/AIDriveStrategyFieldWorkCourse.lua
```diff
@@ -582,12 +596,17 @@ function AIDriveStrategyFieldWorkCourse:getTurnEndForwardOffset()
...
+function AIDriveStrategyFieldWorkCourse:rememberWaypointToContinueFieldWork()
+    local ix = self:getBestWaypointToContinueFieldwork()
+    self.vehicle:rememberCpLastWaypointIx(ix)
+end
+
+function AIDriveStrategyFieldWorkCourse:getRememberedWaypointToContinueFieldWork()
+    return self.vehicle:getCpLastRememberedWaypointIx()
 end
 
 function AIDriveStrategyFieldWorkCourse:getBestWaypointToContinueFieldwork()
```
pvaiko, line 608: W

scripts/ai/AIDriveStrategyFieldWorkCourse.lua
```diff
@@ -582,12 +596,17 @@ function AIDriveStrategyFieldWorkCourse:getTurnEndForwardOffset()
...
-    local bestKnownCurrentWpIx = self.ppc:getLastPassedWaypointIx() or self.ppc:getCurrentWaypointIx()
+    local bestKnownCurrentWpIx = self.course:getLastPassedWaypointIx() or self.course:getCurrentWaypointIx()
```
pvaiko, line 609: Not now, but we'll need to keep the fieldwork course separate and track the progress on that, as self.course is dynamic and may contain any temporary course as well.

pvaiko (changes requested):

scripts/ai/AIDriveStrategyFieldWorkCourse.lua
pvaiko: Oh, just saw your commit message, so course is now always the fieldwork course and we set the temporary course on the ppc only.

schwiti6190: @pvaiko Simply guessed that it might work. Do you think this is good solution ?

pvaiko: As long as we agree that self.course is never changed after the work is started, sure. But then we have to make sure that all functions using the current course (like the onWaypointChange) use the actual course PPC is driving and not self.course.

PR #572, opened 2022-01-17
also need for the Combine strawchopper on headland Turns

RealNeo80:
Combine also needs to wait on the headland turns, that it is not spreading out the straw and grassland or roads


Tensuko:
Did you rly try that ?

RealNeo80:
yep, why? I harvested 5 fields now with this configuration ;)

Tensuko:
So before he starts that turn he discharges the straw ?
But is that rly needed ? I think only on 2nd and more headland since on 1st headland we just drive backwards anyway 1st.

RealNeo80:
yes its standing like on the non-headlands and discharging the straw

needed, in my opinion yes, cause if you drive with a baler afterwards its not easy to drive the curves ;) 
and you have no straw outside the field

Tensuko:
Will not be merged.
Reason:
Combine keeps stopping in the headleand turn, wich makes the turn weird and slow.
There is no benefit for a but straw in a better line.

PR #686, opened 2022-01-25
mini gui update

Bobster82:
Short Summary what I changed

Moved: 'openHudWithMouse' to global settings, as imo this is a global setting and not vehicle depended.

CpHud.lua:
CpHud:onRegisterActionEvents:
	removed: 'if g_Courseplay.globalSettings.controllerHudSelected:getValue() then return end', 
		you still want to use the mouse if mini hud is active in settings.
	Used the global openHudWithMouse setting to register events

Split CpHud:actionEventMouse() into 2 funcs (openCpHud) to use it even when CP_TOGGLE_MOUSE never fires due other mods

getCpHud()
	Added check if active, prevents mouse actions when not active

onUpdateTick()
	Added opening hud when mouse is active and global settings==true (when CP_TOGGLE_MOUSE never fires due other mods)

Changed: Vehicle depended isHudActive instead of on/off for every vehicle

*This changes are all working with and without AD, I did it all by my knowledge as I am no programmer and still learning ;)

### DIFF
--- a/config/GlobalSettingsSetup.xml
+++ b/config/GlobalSettingsSetup.xml
@@ -36,6 +36,7 @@
 	<!-- TODO: save the user settings in the game settings and not the save game dir.-->
 	<SettingSubTitle prefix="true" title="userSettings">
 		<Setting classType="AIParameterBooleanSetting" name="controllerHudSelected" defaultBool="false" onChangeCallback="onHudSelectionChanged" isUserSetting="true"/>
+		<Setting classType="AIParameterBooleanSetting" name="openHudWithMouse" defaultBool="false" onChangeCallback="onHudSelectionChanged" isUserSetting="true"/>
 	</SettingSubTitle>
 	<SettingSubTitle prefix="true" title="pathfinder">
 		<Setting classType="AIParameterSettingList" name="maxDeltaAngleAtGoal" min="5" max="90" incremental="5" default="45"/>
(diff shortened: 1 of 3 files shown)


Tensuko:
> Moved: 'openHudWithMouse' to global settings, as imo this is a global setting and not vehicle depended.

No, it have to stick to the vehicle!





> CpHud:onRegisterActionEvents:
> removed: 'if g_Courseplay.globalSettings.controllerHudSelected:getValue() then return end',
> you still want to use the mouse if mini hud is active in settings.
> Used the global openHudWithMouse setting to register events

Not sure what you want to say with that...

Our current state works fine with AD.
Not sure what all those changes are for...

Bobster82:
>No, it have to stick to the vehicle!
Ok, seems to me a global setting ;)

>Our current state works fine with AD.
>Not sure what all those changes are for...
>We even can use both cursors for AD and CP so no Issue at all.
I know I had issues with opening the mini hud when I changed the setting to open hud with delete key.

I tried, apparently I failed...

Tensuko:
Del key works when mouse is turned off in settings.
Right now we don't support both at the same time, but that can be added later.

The old GUI on the right side is made for gamepad so only need keybind for this.

The Setting is per vehicle becouse mostly it is used to turn right click off for vehicles with front loader.
Adding a global setting instead of changeing it to global might be also an option for later.

But right now, should wait for our next Release... Wich also contains the changelog!

Bobster82:
Ok, thanks for the review... back to learning lua and fs then ;)

PR #758, opened 2022-01-29
Proof of concept based on @pvaiko implementation in LS 19. WIP

schwiti6190:
Still needs validation and backwards compatibility.


scripts/Waypoint.lua
```diff
@@ -1854,6 +1854,25 @@ function Course:addWaypointsForRows()
 	self:enrichWaypointData()
 end
 
+--- Gets waypoints without waypoints in between a row start and row end. 
+function Course:removeWaypointsBetweenRows()
```
pvaiko, line 1858: This works fine if turns are only at row ends. But we also have headland turns, so a better solution would be to mark the waypoints of a row during generation, as there we know exactly that this is a row, between to 180 turns.

pvaiko (changes requested):

pvaiko:
Closed by 382afc1ebf4b0b111c1942c8e1d3d668b9e363cc

PR #822, opened 2022-02-01
Update translation_es.xml

KCHARRO:
Minor tweaks, and translation fix


aorek:
@KCHARRO What is the reason for capitalizing words without following spanish orthographic rules? It's just to know it for next time we touch the same translation file.

PR #816, opened 2022-02-01
Update translation_fr.xml

Lactic68:
Better description for the combine Self Unload functionality not working on Multi tools / convoy mode.
Address #804

PR #846, opened 2022-02-03
SelfUnload Not supported in multitools

Lactic68:
More description about the self unload not available in convoy or multitool mode


Lactic68:
This reference the @Tensuko request in this pull request #816

pvaiko (changes requested):
Could you please make sure there is always a space after the dot.

Lactic68 (commented):
Space added after the point.

pvaiko (approved):
(you know you can have multiple files in the same commit?)

PR #876, opened 2022-02-05
Small fixes 2

o0bonde:


Tensuko:
You have again created that file at the wrong place.
Sry like that I won't merge it.
Keep the file ONLY inside the right folder pls.

PR #935, opened 2022-02-09
Vehicle configurations

Tensuko:
Description check pls

### DIFF
--- a/config/VehicleConfigurations.xml
+++ b/config/VehicleConfigurations.xml
@@ -81,8 +81,9 @@ You can define the following custom settings:
     work areas such as a plow.
 
 - modName: Name of the .zip file (without '.zip')
-    In case a Mod have the same .xml file name as a Vanilla vehicle or implement, add the name of the .zip file to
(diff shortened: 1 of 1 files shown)


config/VehicleConfigurations.xml
```diff
@@ -80,6 +80,10 @@ You can define the following custom settings:
     soon as the the front reaches the row start/field edge, avoiding unworked patches with irregular implement
     work areas such as a plow.
 
+- modName: Name of the .zip file (without '.zip')
+    In case a Mod have the same .xml file name as a Vanilla vehicle or implement, add the name of the .zip file to
+    prevent a false entry. Note: Only 2 of the same .xml files can be added.
+
```
schwiti6190, line 86: In case a Mod has the same .xml filename for the vehicle/implement as a default giants vehicle/implement, add the name of the .zip file to prevent a conflicts. For example: "FS22_exampleMod"
 Note: Only 2 of the same .xml files can be added.

schwiti6190 (changes requested):

PR #950, opened 2022-02-10
WIP: created BaleFinder job and code improvements.

schwiti6190:


scripts/ai/jobs/AIJobBaleFinderCp.lua
```diff
@@ -0,0 +1,65 @@
+--- Bale finder job.
+---@class AIJobBaleFinderCp : AIJobFieldWorkCp
+AIJobBaleFinderCp = {
```
pvaiko, line 3: At some point we should really decide if we want everything to prefix with Cp or postfix...

Issue #326, opened 2022-01-02
Poplar biobaler WB-55 lower [BUG_SP]


Drajo123:
**Steps to reproduce**
after generating course and starting work, worker newer lowers biobaler
**Actual behavior**
after generating course and starting work, worker newer lowers or lift biobaler, if it starting with lifted biobaler it wil yust drive course without harvesting, if you manuali before strating course lower biobaler it will harvest
**Expected behavior**
worker lovering and raising bioblaer as neded

**Screenshots**

**Courseplay version**
Version used: 7.0.0.11

**Log**
[log.txt](https://github.com/Courseplay/Courseplay_FS22/files/7799866/log.txt)

**Map**
Haut

**Vehicles/ implements involved**
biobaler WB-55
dutz fahr 8280 ttv

**Relevant script mods**
[log.txt](https://github.com/Courseplay/Courseplay_FS22/files/7799866/log.txt)

PR #976, opened 2022-02-12
Implement controller improvements

schwiti6190:
- Fixed starting point in hud, where balefinder and course job are
  possible.
- Improved the ai inGame menu integration.
- Improved transition between hud and course generator layout.
- Fixed cottom harvester not released at end of the course.
- Fixes a few tools not lowering/raising at all. #326 #405
- Removed the ai limit for cp jobs/helpers.
- Small mp fixes, maybe #989

### DIFF
--- a/scripts/ai/controllers/CutterController.lua
+++ b/scripts/ai/controllers/CutterController.lua
@@ -1,0 +1,23 @@
+--- Raises/lowers the pickup.
+---@class CutterController : ImplementController
+CutterController = CpObject(ImplementController)
+
+function CutterController:init(vehicle, implement)
+    ImplementController.init(self, vehicle, implement)
+    self.cutterSpec = self.implement.spec_cutter
+end
+
+function CutterController:getDriveData()
+	self:disableCutterTimer()
+	return nil, nil, nil, nil
+end
+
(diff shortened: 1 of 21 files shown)


scripts/ai/controllers/CutterController.lua
```diff
@@ -0,0 +1,34 @@
...
+---@class CutterController : ImplementController
+CutterController = CpObject(ImplementController)
+
+function CutterController:init(vehicle, implement)
+    ImplementController.init(self, vehicle, implement)
+    self.cutterSpec = self.implement.spec_cutter
+end
+
+function CutterController:onLowering()
+	self.implement:aiImplementStartLine()
```
pvaiko, line 11: why is this called here again? Didn't we call this for all implements already from the strategy?

scripts/ai/controllers/CutterController.lua
```diff
@@ -0,0 +1,34 @@
...
+end
+
+function CutterController:onRaising()
+	self.implement:aiImplementEndLine()
```
pvaiko, line 15: Same here

pvaiko (changes requested):

PR #984, opened 2022-02-13
Update translation_sv.xml

piplarsson:


translations/translation_sv.xml
```diff
@@ -4,19 +4,18 @@
     <!-- SV translation By piplarsson -->
 		<text name="CP_activated" 																		text="Aktiverad" />
 		<text name="CP_deactivated" 																	text="Avaktiverad" />
-		<text name="CP_automatic" 																		text="auto" />
```
Tensuko, line 7: why remove ?

piplarsson: I'm sorry, I did the translation from an earlier version.  I can of course redo the latest if needed.  otherwise you may be able to add the missing line

piplarsson: correct translation of that line is "automatisk"

PR #219, opened 2021-12-28
missing file __Attributions.md

xno:

PR #994, opened 2022-02-13
Remove reference to Attributions

Thodor12:
Removes the README section on attributions (proposed by #219), instead added translators directly into README


Thodor12:
Currently I do not know if this list is fully up-to-date with the latest PR's (I have not checked all commits, bit too many for that).
Might think of a way using Actions we can detect if a PR is changing translation files and automatically add them to the contributors list (including backwards compatability).

PR #1002, opened 2022-02-14
Renamed wrong translation codes

Thodor12:
"cs" contained Japanese translations, those should be placed in "ja"
"br" contained Portuguese translations, those should be placed in "pt"
"jp" does not exist in ISO_639-1, it's "ja"
"cs" is the ISO_639-1 of the Czech, "cz" which claims to be Czech does not exist

If I'm wrong and FS22 does not follow standard ISO_639-1 rules and uses different codes feel free to disregard this PR

Also added language names as comments to each file because apparently languages codes are hard to look up


Thodor12:
Someone in the modders help section in the FS Discord pointed me that these language codes are custom defined and do not follow traditional ISO standards, so I'll be closing this PR

PR #1003, opened 2022-02-14
Automatic Action to add translation contributors

Thodor12:
This PR adds an Action to the build pipeline that runs whenever translation files are updated through PR's or direct pushes, the code automatically reads out a list of all edited translation files and appends the user who made the push/pr to an internal contributors.json file.
Using this file a Contributors.md template file is built and placed in the project root, from which the README can link through to it.
The action then automatically commits this change as a separate push to either the PR or the branch


Thodor12:
Note: This branch also contains the changes from #1002 because I immediately wanted to target all the correct translation files

Thodor12:
Turned into WIP for now because the Actions cannot make a commit with the secrets.GITHUB_TOKEN because this is bound to the scope of the user who made the PR, so non project members do not have the right to commit, gonna look into this to see if it's possible nonetheless, else a change is required which uses a PAT

Thodor12:
Had to remove the PR functionality, Github runs Actions on the target branch and thus also attempted to commit the changes there which is not possible, I tried to use `pull_request_target` which supposedly was to fix this by adding write permissions but this is not recommend and has security flaws.
The recommended alternative is to only listen to push events on the main branch, changed it to this so it can be reviewed now

Thodor12:
Undid all of my translation file changes as said in #1002 these codes do not follow ISO standards, I've still added comments to denote what language which file is so it's easier to understand

pvaiko:
> On merging this the pipeline is going to kick off updating the contribs and is going to put me as a translator in all files

Which reminds me that the devs also update the translation files every time they add a feature which needs a translation. Would then all the devs show up as translators for every language?

Thodor12:
> > On merging this the pipeline is going to kick off updating the contribs and is going to put me as a translator in all files
> 
> Which reminds me that the devs also update the translation files every time they add a feature which needs a translation. Would then all the devs show up as translators for every language?

No, I have thought about this, if a user is inside contributors.json under the main section then they are automatically exempted from being put in the translator sections

Thodor12:
@pvaiko You updated the Contributors.md file, this is actually auto-generated through the JS script, if you check in .github/scripts/update_contributors you'll actually see a `contributors.json` file there, that file actually contains the main developers, the translation contributors, etc.
Everything the `index.js` script runs it takes the `contributors-template.md`, opens it, modifies the `[[main]]` and `[[translators]]` blocks and then writes it back out to the `Contributors.md` main file, so purely updating that file will mean the next time the JS script runs it will just reset it back.
In order to fully get rid of unwanted translators or modify the developers you want to modify `contributors.json`

pvaiko:
Damn. Did not realize there's an intermediate step with that json file. Thanks.

pvaiko:
@Thodor12 this did not work out, see b96a88b9

Thodor12:
> @Thodor12 this did not work out, see [b96a88b](https://github.com/Courseplay/Courseplay_FS22/commit/b96a88b9ea7017ce68d4abb5ced8cd809a9a2620)

I had already seen that one, didn't think about that. But that had nothing really to do with this PR itself other than I wanted to add the language names to every file seeing as the game uses modified language codes but not ISO

pvaiko:
> But that had nothing really to do with this PR itself 

Just that we are on the same page here: that PR changed the translate files and broke the complete CP UI. No big deal, shit happens, but saying  that "nothing really to do with this PR" sounds a bit sketchy.

Thodor12:
Yeah does sound a bit sketchy and sorry that the UI broke, didn't anticipate that.
On the good side the action so far seems to be functional, if there's any problems with it you can reach out to me!

PR #1050, opened 2022-02-19
Update translation_jp.xml

Erix058:
Translated to Japanese


Tensuko:
The Translator list and stuff should update automatically, can you pls remove that from your PR ?

PR #1078, opened 2022-02-22
Update contributor-update.yml

Thodor12:


Thodor12:
@pvaiko I disabled "since_last_remote_commit" so the check only runs on the last sent push. This check was only necessary for when I still wanted PR support but that has already been dropped making this setting obsolete. This should fix #1077 (though the translation files have to be reset in that PR)

PR #1109, opened 2022-02-26
updated translation_da

o0bonde:
but why does it allso change [Contributors.md] and contributors.json?


Tensuko:
@Thodor12 
Excuse me sir, this seems to be still bugged. :)

I will edit that PR and add the translations so we can get the work from him into main branch...

Tensuko:
Added this directly to Main Branch.
Will close your PR.

Thodor12:
@Tensuko I found out at least why it went wrong, he merged the upstream of the fork back onto the main, which is considered a push on the main branch and thus the pipeline ran, the changed files from the merge contained the translation files so it caused it to trigger as if he edited them.
I'll have to do a bit of research to see if I can prevent this behaviour. If people make their requests through patch branches it should work fine.

PR #1122, opened 2022-02-28
Orchard driver

schwiti6190:


scripts/Course.lua
```diff
@@ -88,6 +88,12 @@ function Course:getFieldPolygon()
 	return self.fieldPolygon
 end
 
+--- Any vine nodes found on the field?
+function Course:hasVineNodesOnTheField()
+	local x, _, z = self:getWaypointPosition(2)
```
pvaiko, line 93: Magic constants are rarely good and they certainly need an explanation.

scripts/Course.lua
```diff
@@ -88,6 +88,12 @@ function Course:getFieldPolygon()
 	return self.fieldPolygon
 end
 
+--- Any vine nodes found on the field?
+function Course:hasVineNodesOnTheField()
```
pvaiko, line 92: This really isn't the business of the Course class, there is some logic outside of it that needs it.

scripts/specializations/CpAIFieldWorker.lua
```diff
@@ -285,7 +285,12 @@ function CpAIFieldWorker:replaceAIFieldWorkerDriveStrategies(jobParameters)
         spec.driveStrategies = {}
     end
     local cpDriveStrategy
-    if AIUtil.getImplementOrVehicleWithSpecialization(self, Combine) and not AIUtil.hasChildVehicleWithSpecialization(self, VineCutter) then
+    local course = self:getFieldWorkCourse()
+    if course and course:hasVineNodesOnTheField() then
```
pvaiko, line 289: Not sure if this is the best way to decide if this strategy is needed, couldn't we use the same logic which decides which course generator parameters to show?

pvaiko (changes requested):

scripts/specializations/CpAIFieldWorker.lua
schwiti6190: Not all vine implements are unique, for example the sprayer, which is has the same as a normal sprayer. Basically we need a solution for only using this strategy, when a vine field is present. Another possiblity would be saving a flag, like "isVineFieldWorkCourse" in the course class.

scripts/Course.lua
```diff
@@ -268,7 +274,25 @@ function Course:enrichWaypointData(startIx)
 			directionChangeFound = true
 		end
 	end
+	--- Trying to recreate the field polygon.
```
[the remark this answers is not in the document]
schwiti6190: After loading a course from storage the fieldpolygon gets lost. We could save these, but that would also slow down the saving overall and is not the best idea from my perspective.

scripts/Course.lua
schwiti6190: Your right this solution is a suboptimal solution, as this assumes the second waypoint is at least on the field.

PR #1164, opened 2022-03-09
Turns off the cutter from harvesters, while waiting for unloading.

schwiti6190:


scripts/ai/controllers/CutterController.lua
```diff
@@ -9,9 +10,22 @@ end
...
+	--- Turns off the cutter, while the driver is waiting for unloading.
+	if self.driveStrategy.isWaitingForUnload and self.driveStrategy:isWaitingForUnload() then 
+		if self.implement:getIsTurnedOn() then 
+			self.vehicle:aiBlock()
+		end
+	end
 	return nil, nil, nil, nil
 end
 
+function CutterController:onLowering()
```
pvaiko, line 22: Are we sure that every transition out of one of the waiting for unload states will trigger this?

pvaiko (commented):
I'm fine with the code, no idea if

scripts/ai/controllers/CutterController.lua
schwiti6190: Not sure if these cover all possibilities. Was hoping you could maybe check this, as the combine state machine is pretty complex and I didn't want to change anything there.

pvaiko: That's why my first implementation for the sprayer was turning on when it started moving, so here, onRaise would set a flag (wasRaised or something) and then in getDriveData, when the state is not waiting for unload, it could call unblock when not turned on.

pvaiko: even better...

pvaiko (approved):

PR #1213, opened 2022-03-15
Update translation_es.xml

KCHARRO:
100% - Sorry for delay, ADSL problems.


Tensuko:
You are always so fast, no need to be sry :)

PR #1268, opened 2022-03-23
Update translation_it.xml

Roby1164:


Tensuko:
Tjat was fast, awesome.

PR #1288, opened 2022-03-25
update Translation_HU

fuggo2:
7.0.1.7


Tensuko:
Buddy this is still not correct...
The file is old and you added 3 files...

fuggo2:
Unfortunately I haven't received any help so far on how to upload it, so I'm just trying because I don't know how github works. I made the translation from 7.0.1.5 on my own machine in the original translation_hu.xml file. I thought someone would help because as I first wrote I am neither a developer nor a programmer. well if not good then i'm sorry please delete this pullrequest and i won't try again because it doesn't work for me ;) 

so i still think they did a great job with the CP :) i love using it

PR #1345, opened 2022-04-07
Fix automatic translation contributor workflow

Thodor12:
Finally managed to get this fixed, the way this works right now is it reads all the commits that have been pushed, this could potentially include commits coming from merges.
The commits which are not made by the current authenticated Github account are filtered out, after that I use @actions/exec to execute git log on the commits which are made by the current user to check which files have changed.
After that I run the original node command which takes all the input files and makes the contributor update.


pvaiko:
Is there any way to verify this before merging?

Thodor12:
@pvaiko Yes, I just made a test commit to merge the changes of main back into my branch:
https://github.com/Thodor12/Courseplay_FS22/actions/runs/2180101721
In this merge back there have been plenty of other changes that edited translation files, but none of them actually impacted all of the translation files, I only made 2x Test commits that edited the BR translations.
The update contributors commit then added me into the BR translators, but not any of the others.

PR #1362, opened 2022-04-11
Update translation_es.xml

KCHARRO:
Translate by parts, the PC is giving black screens, I think it's the hdd.


Tensuko:
Awesome! Thank you! :)

PR #1372, opened 2022-04-12
Update translation_it.xml

Roby1164:
I have done the translation, but it is not clear what these two items refer to, as they do not refer to a precise level? In fact I put "Low" as an indication, in order to have a point of reference, thinking that you mean this type of information.


KCHARRO:
It must be to set a minimum limit, so that CP stops or sends to AD, to the workshop, or to refuel(as it was in FS19.)

Tensuko:
Yes, exactly like @KCHARRO said.

PR #1379, opened 2022-04-13
Multitool course start fix for #1347

pvaiko:
The offset course must be calculated when the drive to fieldwork
task starts so it can find and drive to the nearest/first waypoint
of the _offset_ course, not the original course.

### DIFF
--- a/scripts/ai/AIDriveStrategyCourse.lua
+++ b/scripts/ai/AIDriveStrategyCourse.lua
@@ -128,9 +128,31 @@ function AIDriveStrategyCourse:delete()
     AIDriveStrategyCourse:superClass().delete(self)
 end
 
-
 function AIDriveStrategyCourse:getGeneratedCourse(jobParameters)
-    return self.vehicle:getFieldWorkCourse()
+    local course = self.vehicle:getFieldWorkCourse()
+    local numMultiTools = course:getMultiTools()
+    local laneNumber = numMultiTools > 1 and jobParameters.laneOffset:getValue() or 0
+    if numMultiTools < 2 then
+        self:debug('Single vehicle fieldwork course')
+        self.vehicle:setOffsetFieldWorkCourse(nil)
+        return course
+    elseif laneNumber == 0 then
+        self:debug('Multitool course, center vehicle, using original course')
+        self.vehicle:setOffsetFieldWorkCourse(nil)
+        return course
+    else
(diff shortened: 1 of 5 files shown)


Tensuko:
Works fine, if @schwiti6190 reviewed the code, this can be merged.

scripts/ai/AIDriveStrategyCourse.lua
```diff
@@ -128,9 +128,31 @@ function AIDriveStrategyCourse:delete()
     AIDriveStrategyCourse:superClass().delete(self)
 end
 
-
 function AIDriveStrategyCourse:getGeneratedCourse(jobParameters)
-    return self.vehicle:getFieldWorkCourse()
+    local course = self.vehicle:getFieldWorkCourse()
```
schwiti6190, line 132: A simple nil check, if a course was found might be needed for the bale finder.

scripts/specializations/CpCourseManager.lua
```diff
@@ -204,14 +206,17 @@ function CpCourseManager:setCoursesFromNetworkEvent(courses)
...
+    spec.offsetFieldWorkCourse = nil
     course:setVehicle(self)
     table.insert(spec.courses,course)
     SpecializationUtil.raiseEvent(self,"onCpCourseChange",course,noEventSend)
 end
 
 function CpCourseManager:resetCourses()
-    local spec = self.spec_cpCourseManager 
+    local spec = self.spec_cpCourseManager
+    spec.offsetFieldWorkCourse = nil
```
schwiti6190, line 219: These resets are only applied, when the course changes. This might be a problem, if the user want to start the course with different lane offsets. For example the wrong lane offset was used and then restarted with a different lane offset.

schwiti6190 (changes requested):

scripts/specializations/CpCourseManager.lua
pvaiko: That is taken care of elsewhere, that is why we save the offsetFieldWorkPosition.

scripts/ai/AIDriveStrategyCourse.lua
pvaiko: Good catch.

scripts/specializations/CpCourseManager.lua
schwiti6190: Seems like I over looked this: "position ~= previousPosition" check. Might be okay then, but a test with AD and refill would still be usefull.

PR #1405, opened 2022-04-16
Replacing tabs with spaces

KITT3000:
Replacing tabs with spaces, FS doesn't like tabs -> `Warning: Character '9' not found in texture font`


pvaiko:
Thanks!

PR #1406, opened 2022-04-17
Update translation_es.xml

KCHARRO:
Late..., I'm on vacation for a week an half, it may take a while to review and update, every two or three days.


Tensuko:
No problem

Issue #1409, opened 2022-04-17
[BUG_MP] Speed settings no longer work


Selyria:
Steps to reproduce

The speeds can be set but are not adopted since version 7.0.1.12

Speeds are not taken over since version 7.0.1.12

Neither slower nor faster

Actual behavior

With version 7.0.1.11 my cultivator could still drive 32 KM/H with version 7.0.1.12 only 25 although I have switched on the expert mode and have the speed on 40 stand

Expected behavior

It should be able to reach the Max KM/H also

Screenshots

https://youtu.be/gH9-Y5S_yC0

Courseplay version
Version used: 7.0.1.12

Client Log

[log.txt](https://github.com/Courseplay/Courseplay_FS22/files/8501465/log.txt)

Server Log

[log.txt](https://github.com/Courseplay/Courseplay_FS22/files/8501467/log.txt)

Player on the server
4

Dedicated server or local hosted server
Dedicated server Yes

Map

Nordfriesische Marsch 4fach] 1.5

Vehicles/ implements involved

Kein Mod Fahrzeug

Relevant script mods
Autodrive 2.0.0.5

Tensuko:
The settings might work again correctly now, pls test and reopen if Issue persists.

PR #1414, opened 2022-04-17
Simple mode only returns the default values in Singleplayer.

schwiti6190:
#1409


Tensuko:
Tested with a course generator setting, is that enough ?

schwiti6190:
@Tensuko Vehicle settings test would be better, as the course generator settings are only local anyway.

PR #1421, opened 2022-04-19
Ability to skip 2 lines in vines

jsalatas:
This just a quick edit to give the ability to skip two lines in vine course. 

Useful for the Marshall ST1800 slurry spreader mod (https://www.farming-simulator.com/mod.php?lang=en&country=us&mod_id=233075&title=fs2022)

Screenshot of the mod working in an olive grove


Tensuko:
Becouse it works on 4 rows and not only 2 ?

jsalatas:
> Becouse it works on 4 rows and not only 2 ?

Yes! Exactly!

PR #1439, opened 2022-04-21
Update translation_it.xml

Roby1164:
correction for incorrect string, no title but text   "<text name="CP_help_page_GeneralBase_text"

### DIFF
--- a/translations/translation_it.xml
+++ b/translations/translation_it.xml
@@ -349,25 +349,25 @@
 		<!--This stays blank and is just to create blank spots for better overview in the menu-->
 		<text name="CP_help_page_blank"																	text=""/>
 		<!--General-->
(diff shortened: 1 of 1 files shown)


Tensuko:
I might was to fast with your PR then, now we have a conflict, can you check to resolve it?

Roby1164:
This string is incorrect in all translations

Roby1164:
with the wording "Title", gives an error, the string must report "Text"
Incorrect: "<text name="CP_help_page_GeneralBase_title"
Correct: "<text name="CP_help_page_GeneralBase_text"

Tensuko:
I will check that, thanks for pointing it out.

Tensuko:
I corrected this for all translations.
Thanks for noticeing it!

PR #1465, opened 2022-04-24
Issue 480

pvaiko:
Work/width offset fix

Auto work width/offset calculation reworked. Now using the left and right distances over all implements (instead of just the width) and calculate final working width/offset from that.

Configured width/offset always takes precedence over automatically detected values.

### DIFF
--- a/scripts/specializations/CpCourseGeneratorSettings.lua
+++ b/scripts/specializations/CpCourseGeneratorSettings.lua
@@ -85,8 +85,7 @@ end
 function CpCourseGeneratorSettings:onLoadFinished(savegame)
     local spec = self.spec_cpCourseGeneratorSettings
     if not spec.wasLoaded then
-        local width, offset WorkWidthUtil.getAutomaticWorkWidthAndOffset(self)
-        spec.workWidth:setFloatValue(width)
+        CpCourseGeneratorSettings.setAutomaticWorkWidthAndOffset(self)
     end
 end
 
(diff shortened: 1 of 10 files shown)


scripts/specializations/CpCourseGeneratorSettings.lua
```diff
@@ -174,11 +164,6 @@ function CpCourseGeneratorSettings:raiseCallback(callbackStr, setting, ...)
     SpecializationUtil.raiseEvent(self, callbackStr, setting, ...)
 end
 
-function CpCourseGeneratorSettings:setAutomaticWorkWidth()
```
schwiti6190, line 177: This function was also use for the hud setting: ``
<Setting classType="AIParameterSettingList" name="workWidth" min="1" max="50" incremental="0.1" unit="2" setDefault="setAutomaticWorkWidth" onChangeCallback="cpShowWorkWidth"/>
``

schwiti6190: Might be a good idea to move these code parts into this function:
 `
local spec = self.spec_cpCourseGeneratorSettings
local width, offset = WorkWidthUtil.getAutomaticWorkWidthAndOffset(self)
spec.workWidth:setFloatValue(width)
`

schwiti6190 (changes requested):

scripts/specializations/CpCourseGeneratorSettings.lua
pvaiko: this was the fix for clicking width/offset in the HUD, should now be working fine.

PR #1582, opened 2022-05-22
Unloading StonePicker

good49:
Can you Update Courseplay for unloading Stone with "Wheel loader / Telehandler" and stone picker.

project :
The StonePicker as a capalities harvester.
With Autodrive the wheel loader unloading the stonepicker and return to bunker or other to unloading stone.

exampleunloading


Tensuko:
Can you NOT make this request as a super weird PullRequest ?
Thanks.

PR #1629, opened 2022-05-28
Automatically changes the sprayer workwidth, when the filltype changed.

schwiti6190:
TODO: Figure out when to correctly apply this, as a possible manual work width will be discarded on a filltype change with a sprayer attached.


Tensuko:
I would say like this, its totaly fine.
As soon as you load a filltype, the work width will be set.
If you change it manually and fill it up again, it will stay manual value.
Still can use autoset.

Actually it is just to not have to press the autoset work width so I guess we have improoved it.

PR #1649, opened 2022-05-31
Update translation_ru.xml

Gonimy-Vetrom:
Hello. I'm still alive. After three months of the war, I had the Internet in a new place. I don't know-how for a long time. Therefore, so far made part of the translation. And then that I will have time to send.
When are you planning to make a release? Maybe I'll try to hurry.


Tensuko:
@Gonimy-Vetrom 
Hey buddy.
Well we have to do a hotfix release now, but will add your first changes ofc.
Before ModHub, there might be another Version first, as we need to do some big MP testing.

Issue #1679, opened 2022-06-05
Translation to Spanish


KCHARRO:
If anyone notices any untranslated text in Spanish, please let me know.

Tensuko:
Line 10
Line 38 - 42
Line 53 and 54
Line 115 and 116
Not sure about Line 210
Line 363 maybe

As far as I can tell, thats all :)

KCHARRO:
Done....

PR #1685, opened 2022-06-06
Update translation_es.xml

KCHARRO:
All done.


KCHARRO:
Finished everything mentioned.

on line 210, the word "Manual" is spelled the same and means the same in both languages, for example:

Trabajo Manual
Manual Work (here can also be: Handwork)

Tensuko:
And #1679 done :)

PR #1851, opened 2022-08-11
Combine unloader hud

schwiti6190:
- Added new HUD buttons for the combine unloader.
- Gamepad hud still TODO
- Drive now button needs to implemented.
- Tooloffset z also needs to be implemented.
- Figure out a good way to show the tooloffsets for the trailer, 
  separate from the normal workwidth change.

### DIFF
--- a/config/VehicleSettingsSetup.xml
+++ b/config/VehicleSettingsSetup.xml
@@ -57,6 +57,9 @@
 		<Setting classType="AIParameterSettingList" name="toolOffsetX" min="-10" max="10" incremental="0.1" default="0" unit="2"
 				 onChangeCallback="cpShowWorkWidth" setDefault="setAutomaticWorkWidthAndOffset"
 				 isDisabled = "isToolOffsetDisabled" vehicleConfiguration="toolOffsetX"/>
+		<!--Tool Offset Z-->
+		<Setting classType="AIParameterSettingList" name="toolOffsetZ" min="-10" max="10" incremental="0.1" default="0" unit="2"
+				 onChangeCallback="cpShowWorkWidth" isDisabled = "isToolOffsetDisabled" />
 		<!--Silage additives needed?-->
 		<Setting classType="AIParameterBooleanSetting" name="useAdditiveFillUnit" defaultBool="false" isVisible="isAdditiveFillUnitSettingVisible" isExpertModeOnly="true"/>
 	</SettingSubTitle>
--- a/scripts/ai/jobs/CpAIJobCombineUnloader.lua
+++ b/scripts/ai/jobs/CpAIJobCombineUnloader.lua
@@ -3,7 +3,7 @@
 CpAIJobCombineUnloader = {
 	name = "COMBINE_UNLOADER_CP",
 	translations = {
-		jobName = "CP_job_unloadCombine"
+		jobName = "CP_job_combineUnload"
 	}
 }
 local AIJobCombineUnloaderCp_mt = Class(CpAIJobCombineUnloader, CpAIJobFieldWork)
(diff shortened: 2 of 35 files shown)


scripts/specializations/CpAICombineUnloader.lua
```diff
@@ -162,3 +163,8 @@ function CpAICombineUnloader:replaceDriveStrategies(fieldPolygon, jobParameters)
     --- Only the last driving strategy can stop the helper, while it is running.
     table.insert(spec.driveStrategies, cpDriveStrategy)
 end
+
+--- Forces the driver to unload now.
+function CpAICombineUnloader:startCpCombineUnloaderUnloading()
+    print("Drive now!")
```
schwiti6190, line 169: @pvaiko This is the button callback from drive to unload (driveNow), that needs to be implemented.

scripts/ai/jobs/CpAIJobCombineUnloader.lua
```diff
@@ -83,6 +82,7 @@ end
 
 --- Called when parameters change, scan field
 function CpAIJobCombineUnloader:validate(farmId)
```
schwiti6190, line 84: @pvaiko Not sure why you didn't use the same setup, as with the bale finder?

config/VehicleSettingsSetup.xml
```diff
@@ -57,6 +57,9 @@
 		<Setting classType="AIParameterSettingList" name="toolOffsetX" min="-10" max="10" incremental="0.1" default="0" unit="2"
 				 onChangeCallback="cpShowWorkWidth" setDefault="setAutomaticWorkWidthAndOffset"
 				 isDisabled = "isToolOffsetDisabled" vehicleConfiguration="toolOffsetX"/>
+		<!--Tool Offset Z-->
+		<Setting classType="AIParameterSettingList" name="toolOffsetZ" min="-10" max="10" incremental="0.1" default="0" unit="2"
```
schwiti6190, line 61: @pvaiko Vertical offset(z)

scripts/ai/jobs/CpAIJobCombineUnloader.lua
```diff
@@ -63,16 +62,14 @@ function CpAIJobCombineUnloader:applyCurrentState(vehicle, mission, farmId, isDi
 		end
 	end
 
+	x, z = vehicle:getCpCombineUnloaderFieldPosition()
```
pvaiko, line 65: @schwiti6190 why is this different/better than the code segment above with lastJob?

scripts/ai/jobs/CpAIJobCombineUnloader.lua
pvaiko: Not sure what you mean. With the unloader, we want two positions: 
- one **outside** of the field where the Giants AI should drive it, and where it waits for the combine
- one on the field, which we are not driving to, only use it to mark the field we are working on. 
So to me, this is different than the bale finder.

schwiti6190: The same two positions are used for the balefinder, when the mode is started from the GUI.
The only difference might be the position of the giants marker.

The only meaningful difference between balefinder and unloader is the starting from the HUD, as we don't want to start on the field(balefinder) but in front of it. I've corrected this for the unloader, so the hud isn't reseting the position.

scripts/ai/jobs/CpAIJobCombineUnloader.lua
schwiti6190: The idea is to save the position of the field marker into the vehicle(job), so it can be saved and reloaded.
The in game menu works with blank new Jobs, everytime it's opened. 
This deletes the selected field position on reopening of the job.
The fieldwork/balefinder jobs allways use the current vehicle position as reference for the field selection.

PR #1884, opened 2022-08-31
Only show combine unloader job when only 1 Trailer is attached to the unloader (and weight implementation)

tn4799:
Since multiple trailers aren't supported the job shouldn't get shown if more than one trailer-spec is part of the unloader and it`s attached trailers. After this rework this is realized.

I've also added a AIUtil-function to get the amount of specs that are in such a chain.

Additionally the weight implementation is also a part of this because ive created the new branch at a wrong point.


scripts/ai/FillLevelManager.lua
```diff
@@ -164,6 +164,7 @@ function FillLevelManager:getAllFillLevels(object, fillLevelInfo)
             if not fillLevelInfo[fillType] then fillLevelInfo[fillType] = {fillLevel=0, capacity=0} end
             fillLevelInfo[fillType].fillLevel = fillLevelInfo[fillType].fillLevel + fillUnit.fillLevel
             fillLevelInfo[fillType].capacity = fillLevelInfo[fillType].capacity + fillUnit.capacity
+            fillLevelInfo[fillType].weightLimitReached = g_currentMission.missionInfo.trailerFillLimit and object:getMaxComponentMassReached()
```
schwiti6190, line 167: Old code

tn4799: removed

PR #1936, opened 2022-09-26
Update translation_it.xml

Roby1164:


Tensuko:
We have update this: `<text name="CP_vehicle_setting_fullThreshold_tooltip" text="TODO"/>`
To this: `<text name="CP_vehicle_setting_fullThreshold_tooltip" text="If the unloader's fill level is over this value, it won't wait for the combine"/>`
If you update your one and want to translate it, I can merge it :)

PR #1978, opened 2022-10-05
Ignore sprayer fill level Check if capacity is 0 in sowing maschine

tn4799:
If a sowing machine has a fertilizer capacity of 0 it should just work without calling for a refill when use fertilizer is active. 
Ive added a check if the total capacity of the sprayer fillType is 0. If that is the case the check of the sprayerFilllevel <= 0 will be ignored.

Found out in #1975


scripts/ai/FillLevelManager.lua
```diff
@@ -37,16 +37,17 @@ function FillLevelManager:debugSparse(...)
     end
 end
 
-function FillLevelManager:needsRefill()
+function FillLevelManager.needsRefill()
```
pvaiko, line 40: This function isn't used anywhere, obsolete, can be removed.

scripts/ai/FillLevelManager.lua
```diff
@@ -37,16 +37,17 @@ function FillLevelManager:debugSparse(...)
...
-    return self:areFillLevelsOk(fillLevelInfo)
+    FillLevelManager.getAllFillLevels(self.vehicle, fillLevelInfo)
+    return FillLevelManager.areFillLevelsOk(fillLevelInfo)
 end
 
 
 -- is the fill level ok to continue? With fillable tools we need to stop working when we are out
 -- of material (seed, fertilizer, etc.)
-function FillLevelManager:areFillLevelsOk(fillLevelInfo,isWaitingForRefill)
+function FillLevelManager.areFillLevelsOk(fillLevelInfo, isWaitingForRefill)
```
pvaiko, line 49: Not used, can be removed. Also, helperBuysFillType()

scripts/ai/FillLevelManager.lua
```diff
@@ -104,7 +105,7 @@ function FillLevelManager:areFillLevelsOk(fillLevelInfo,isWaitingForRefill)
 end
 
 --- Does the helper buy this fill unit (according to the game settings)? If yes, we don't have to stop or refill when empty.
-function FillLevelManager:helperBuysThisFillType(fillType)
+function FillLevelManager.helperBuysThisFillType(fillType)
```
pvaiko, line 108: Not used, can be removed.

scripts/ai/FillLevelManager.lua
```diff
@@ -154,30 +155,32 @@ end
...
-            local fillTypeName = g_fillTypeManager:getFillTypeNameByIndex(fillType)
-            self:debugSparse('%s: Fill levels: %s: %.1f/%.1f', object:getName(), fillTypeName, fillUnit.fillLevel, fillUnit.capacity)
-            if not fillLevelInfo[fillType] then fillLevelInfo[fillType] = {fillLevel=0, capacity=0} end
-            fillLevelInfo[fillType].fillLevel = fillLevelInfo[fillType].fillLevel + fillUnit.fillLevel
-            fillLevelInfo[fillType].capacity = fillLevelInfo[fillType].capacity + fillUnit.capacity
-            --used to check treePlanter fillLevel
-            local treePlanterSpec = object.spec_treePlanter
-            if treePlanterSpec then
-                fillLevelInfo[fillType].treePlanterSpec = object.spec_treePlanter
+        for index, fillUnit in pairs(object:getFillUnits()) do
```
pvaiko, line 161: _ in Lua (and in Pyhton) is used for throw away variables we don't care about. No need to put 'index' here, decent IDEs will only complain that it is unused.

scripts/ai/FillLevelManager.lua
```diff
@@ -28,7 +28,7 @@ function FillLevelManager:init(vehicle, debugChannel)
 end
 
 function FillLevelManager:debug(...)
-    CpUtil.debugVehicle(self.debugChannel, self.vehicle, ...)
+    CpUtil.debugVehicle(self.debugChannel or CpDebug.DBG_IMPLEMENTS, self.vehicle, ...)
 end
 
 function FillLevelManager:debugSparse(...)
```
pvaiko, line 34: Should use CpUtil.debugVehicle() with an object.
@schwiti6190 : we should consider changing CpUtil.debugVehicle() to add the rootVehicle to the logging if vehicle ~= rootVehicle.

scripts/ai/FillLevelManager.lua
```diff
@@ -28,7 +28,7 @@ function FillLevelManager:init(vehicle, debugChannel)
 end
 
 function FillLevelManager:debug(...)
```
pvaiko, line 30: obsolete

pvaiko (changes requested):
I have no problem making FillLevelManager a utility class with no data members as it is the only way it is used. 

Then we can remove init() and the instantiation in AIDriveStrategyCourse, which is never used either.

tn4799:
Should I move the FillLevelManager into the util-folder since it is now a utils-table?

pvaiko:
> Should I move the FillLevelManager into the util-folder since it is now a utils-table?

No, util is for more generic utilities.

PR #1989, opened 2022-10-08
Better turning radius for vanilla auger wagons

MeRida2104:


Tensuko:
Thanks but this won't be needed for the next version.
We are implementing a minimum default radius of 10 for all auger wagons.

MeRida2104:
Ok, much better solution :)

PR #2002, opened 2022-10-12
Update translation_cz.xml

SniperKittenCZ:


Tensuko:
Sry we pushed some Updates now your PR got a conflict.
Can you check for it and maybe also add in the new Translations ? :)

Issue #2007, opened 2022-10-13
[BUG_SP] Unloading into american trailers


Misamangame:
Hello guys,

I seen few threads about unloader problems. For me tractors with auger wagons will not unload into vanilla american trailers. I would like to know if someone else experienced this, maybe I'm doing something wrong? Auger wagons unfold pipe and then end unloading process and disengage. If I swap this american trailer for normal one towed with tractors ( I tried 3 vanilla ones) all works without  problems. Logfile is attached, will be a bit mess as I had at that moment lots of machines running also 4 other unloaders with normal trailers for sugarbeet harvesters, apart from 3 with auger wagons for combines mentioned. Trucks are AD driven. Short video here : https://youtu.be/_2iUKNqa6vU

   
![20221013132144_1](https://user-images.githubusercontent.com/33344465/195615674-27e24d34-a9fd-46ec-9adb-14152a7294c9.jpg)

**Courseplay version**
Version used: 7.1.1.4

**Log*
[log.txt](https://github.com/Courseplay/Courseplay_FS22/files/9777230/log.txt)

**Map**
Green Valley Nebraska 4x

pvaiko:
If you turn CP off and open the pipe at the same position, will it start discharging?

rohne83:
Think the cause for the issue here might be a "bad" combination of grain cart's pipe discharge/raycast node and the trailers fill node. So auger wagon thinks it cannot unload as the pipe is "aiming" outside of the trailer.
Saw this happen from time to time also with completely different (more european ones) combinations of auger wagons and transport trailers as well.
Not sure what might be the best solution here, whether user sets the desired pipe position manually (as was possible in FS19) or CP tries to adjust itself.

Misamangame:
@pvaiko @rohne83 @Tensuko update : 

No it will not discharge in that position where the tractor is ending up. But this is case of Bergmann & Hawe unloaders and Wilson trailer. Bergmann can unload to Lodeking to part which is empty (1st, 2nd or 3rd) but then also CP disengage when he fills up only that 1 part. I saw some video where unloader reversed, but it is not a case here.  with HAWE i had same result like with Wilson trailer only difference he managed to discharge few litres before CP disengaged.

I guess what @rohne83 is saying is correct I think generally all auger wagons should line up a bit further away from trailers as they seems way too close. 

I can also test american augers later

Tensuko:
Further away wont work for some. This is testing experience and trust me I tested a lot...
[...]

Issue #2006, opened 2022-10-13
[BUG_SP] Unloader not continuing the work, after unloading into Trailer (colission?)


manierim:
**Steps to reproduce**
Harvester running a CP course, 
Tractor with Auger Wagon running CP unload combine.
AD truck with trailer parked either inside or close to the field border.

**Actual behavior**
The CP unloader tractor drives to the AD trailer and unloads the auger wagon, then it just stays there - also when the combine is full and waiting for unload. 

If the CP unload work is stopped and then reactivated without moving the tractor is still stays there. If the tractor is moved a little bit away from its unload position (or if the AD drives moves to unload) then the CP unloader will finally start moving toward the combine as expected.

**Expected behavior**
The CP unloader should start moving toward the combine after unloading into the waiting AD trailer.

**Screenshots**
![fsScreen_2022_10_13_15_24_21](https://user-images.githubusercontent.com/2349481/195609776-53c8c5cb-a802-4af1-9e15-294c96c349b9.png)

**Courseplay version**
Version used: 7.1.1.4

**Log**

https://gist.github.com/manierim/6f9d83756a1f084b0a09ba9dd7023756

Looks like the problem is detected collisions with the trailer/truck:

```
2022-10-13 15:19 :50 [dbg7 lp194005] pathfinder colliding shapes  TRANSFORM_GROUP SHAPE with Fastrac 8330 at x = 863.5, z = -287.5, (1.5x3.8), yRot = -7
2022-10-13 15:19 :50 [dbg7 lp194005] collision: Anthem 6x4 (MP8 505)
2022-10-13 15:19 :50 [dbg7 lp194005] collision: SKS 30/1050
2022-10-13 15:19 :50 [dbg7 lp194005] pathfinder colliding shapes  TRANSFORM_GROUP SHAPE with GTW 330 at x = 864.0, z = -295.0, (2.6x5.8), yRot = -1
2022-10-13 15:19 :50 [dbg7 lp194005] collision: Anthem 6x4 (MP8 505)
2022-10-13 15:19 :50 [dbg7 lp194005] collision: SKS 30/1050
```

**Map**
Elmcreek

**Vehicles/ implements involved**
Harvester X9 1100
Unloader Fastrac 8330 + Auger GTW 330
AD unload target: Anthem 6x4 + SKS 30/1050

**Relevant script mods**
AutoDrive  2.0.0.7

Issue #2001, opened 2022-10-12
[Feature] Unloader unload in a double Trailer


PapaTim68:
**What is the problem this feature/enhancement solves?**
If you use the new Unloader system with overloading to trailers at the field, it will not work when there are two trailers behind each over connected to one Tractor. The Unloader will endlessly try and fill the first Trailer.

**Describe the solution you'd like**
The unloader recognize that it's two trailers and fill the second when the first is filled.

**Describe alternatives you've considered**
For now, I will refrain from using two trailers and just use one.

Tensuko:
Can't reproduce.

PapaTim68:
Might be again me using the Holmers as Unloaders. Will make some logs and Screenshots later or tomorrow.

PapaTim68:
[log.txt](https://github.com/Courseplay/Courseplay_FS22/files/9767236/log.txt)
https://youtu.be/clq4NCNAehc
Managed to do it faster than expected. Here is a log and a short Video Clip.

Tensuko:
Same thing even when the holmer is further away?

PapaTim68:
Yes I noticed it when the holmer just unloaded from a combine on the field and came back to unload into the Trailers.

pvaiko:
Can you try it with a lower field speed, like 25?

pvaiko:
For the record:
1. The unloader does not have special logic to _continue_ to find another trailer when the first trailer is full but there is fruit left in the unloader. Currently only a self unloading combine can do that.
2. In this issue, the unloader isn't repeatedly trying the first trailer, it is trying the second one but drives so fast that can't stop in time. 
3. The trailer's fill node is in the front when empty and we use that, should probably improve on that and either always use the center of the trailer when unloading or limit how far forward we are willing to drive.

DemsMani:
I seem to have the same problem when trying to load the Super-B Trailers (vanilla game). They are very long as they have in total four compartments and even with a low field - speed (15 km/h) it ever only stopped on the last compartment and is then not able to fill the other three.

rohne83:
don't think this is just the Holmer. As I noted in discord I had same issue as in PapaTims video with the Bergmann beet overloader. And also during my first tests with normal grain carts (Bergmann/HAWE) I haven't managed yet to get the overloader to unload into both trailers.
[...]

PR #2017, opened 2022-10-16
Issue 2001, 2006 and 2007

pvaiko:
- use the discharge node/target node distance for a
    more accurate pipe positioning, auger wagon pipe should
    not overshoot the target anymore, or if it does, it'll
    back up a bit
    - if we can't discharge anymore into a trailer but has
    more than 10% fruit, it'll look for another trailer,
    or another fill target of the same trailer. This
    should improve how long trailers with multiple fill
    nodes are handled
    - once done with unloading, move ahead a bit so the
    pathfinder has a clear view when it needs to drive
    to the combine

#2001 #2006 #2007

### DIFF
--- a/config/VehicleSettingsSetup.xml
+++ b/config/VehicleSettingsSetup.xml
@@ -98,7 +98,7 @@
 	</SettingSubTitle>
 	
 	<SettingSubTitle title="multitool" isVisible="hasMultiToolCourse">
-		<!--Symetric Langechange-->
+		<!--Symmetric Lane Change-->
 		<Setting classType="AIParameterBooleanSetting" name="symmetricLaneChange"/>
 		<!--Convoy Distance-->
 		<Setting classType="AIParameterSettingList" name="convoyDistance" min="40" max="400" default ="75" unit="2" isExpertModeOnly="true"/>
--- a/scripts/ai/AIDriveStrategyCombineCourse.lua
+++ b/scripts/ai/AIDriveStrategyCombineCourse.lua
@@ -114,11 +114,10 @@ end
 function AIDriveStrategyCombineCourse:setAllStaticParameters()
     AIDriveStrategyCombineCourse.superClass().setAllStaticParameters(self)
     self:debug('AIDriveStrategyCombineCourse set')
-  
+
     if self:isChopper() then
         self:debug('This is a chopper.')
     end
-
 
     self:checkMarkers()
     self:measureBackDistance()
@@ -300,7 +299,7 @@ function AIDriveStrategyCombineCourse:driveUnloadOnField()
             local unloaderWhoDidNotShowUp = self.unloaderToRendezvous:get()
             self:cancelRendezvous()
             if unloaderWhoDidNotShowUp then
-                unloaderWhoDidNotShowUp:onMissedRendezvous(self)
+                unloaderWhoDidNotShowUp:getCpDriveStrategy():onMissedRendezvous(self)
             end
             self:debug('Waited for unloader at the end of the row but it did not show up, try to continue')
             self:changeToFieldWork()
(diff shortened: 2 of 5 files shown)


pvaiko:
> After the auger unloader moves forward a bit to avoid collision, it stays in WAITING_FOR_COMBINE_TO_CALL state forever... also if the combine actually stops and await unload.

Yeah, looks like it is still too close, should probably move further ahead. Would you have the full log and a screenshot of the unloader?

manierim:
Ahh no sorry. I'll make a new test and take log and screenshot. By the way there were no collision messages in the console after the unloader moved from the unload position

manierim:
Here it is:

[log.txt](https://github.com/Courseplay/Courseplay_FS22/files/9818641/log.txt)

fsScreen_2022_10_19_10_44_11

I also recorded a [short video](https://drive.google.com/file/d/1gzKDLJagoyZz3vdNcphr2U5nGbSmnSLa/view?usp=sharing) (console opened during unloading)

Tensuko:
A bit more further might help.
fsScreen_2022_10_19_14_04_22

[log.txt](https://github.com/Courseplay/Courseplay_FS22/files/9820230/log.txt)

rohne83:
I see you implemented a threshold until which the grain cart will search for another trailer if it could not empty itself completely. This is absolutely good idea, but I have problems with the exact value. Significantly less than 50% it should be definitely. but 10% is already too low again for practice. Real example from practice with good example values for deriving:
There's just one good beet auger so far which is the Bergmann RRW with capacity of 50k (and the Amity which still can take 45k) which is more than you can transport over roads - especially with activated Trailer Fill Limit you're limited to about 35-40k per vehicle. If the transport vehicles had even less capacity it would definitely make sense for the cart to fill even two of them, but for those I mentioned it should just fill one of them before coing back to the harvesters. Same relations can be applied to much different trailer capacities. This translates into a recommendable value for the threshold of at least 20 or rather even 30%, which I think is a good compromise between 0 and 50%.

manierim:
Commit [aad93c0](https://github.com/Courseplay/Courseplay_FS22/pull/2017/commits/aad93c00214c3c29b20601167f8521f56d86604b) tested well. issue #2006 seems fixed.

rohne83:
> But I dont want more settings. Overall: Make sure we have less fillVolume left untill we are ready to unload a combine again.
> 
> If unloading is too slow and combine have to wait, you got to accept it or add another unloader.
> 
> Same for Transport.

I already have one unloader for each sugarbeet harvester, using even more augers wouldn't solve problems but even create new ones ;)
I'd also prefer the customizable setting, but if you want it fixed then be it. But it should leave some buffer, to allow for smooth operations. 10% would definitely be too strict in that case in very most situations. This will nearly always result in the auger having to unload into multiple transport vehicles, no matter how good you match up your logistics capacities (and I always have a strong focus on that), which is a big problem especially when harvesting fruits with high yield like sugarbeet.

rohne83:
> That would be the case when over 10% (e.g. 11%) left. Below doesnt matter. But that is the same for everything, even if you set it to 30, then it ends up with 31 and you end up in the same situation. 10% is the first try, if 15 fits more its also okay, but I would not go higher.

Well, look at it this way: for efficient use of an auger wagon, you should always try to have your unloaders capacities perfect integer multiples of the auger wagon capacity. So it can always dump its full load into just one single unloader, and doesn't have to circle around about half full. But 1. capacities of the vehicles available in FS don't always match perfectly (especially with the potatoe and sugarbeet carts), and 2. even if they do, masses of the fruits still vary a bit, which doesn't have impact on the grain carts but reduces capacities of the normal AD unloaders.
I.e. while for barley you can transport about 40k with one AD truck and thus you might go for a grain cart with capacity of 38 or 43k, for sorghum truck capacity goes down to about 33k due to the higher mass. Leaving your grain cart with up to 10000 liters in its trailer after overloading = fill level of a bit more than 20% remaining. That's absolutely enough capacity (33000liters) to go back to the combines and empty 2 or 3 of them, instead of staying at the field edge for overloading another just 10000 liters (and then your logistics don't match anymore at all, as the trailer just filled now only has capacitiy left of 23k - not worth to go to the silo, but not enough for getting the next cart unloaded to a really decent level).
With a less strict threshold you can avoid those issues in many situations, but with 10% you'll nearly always have the situation that the grain cart needs a second unloader.
Also with much different grain carts the relations remain quite comparable. The smallest ones available have a capacity of 19k (means 10% allows a mismatch of trucks and grain cart capacity of just 1900liters, while even with 30% threshold it would always have more than enough capacity left to unload another combine). And the largest ones have around 70k.

Tensuko:
We do it now as we planed.

Issue #2045, opened 2022-10-27
[BUG_SP] Move entry will delete whole folder when double click it or hit action key


bojanh66:
I have tried to move some course from folder X to folder Y in game.
I selected the course from folder X, pressed move entry, then selected another folder Y to move the course and now my folder Y will all courses is gone.
I have tried this several times.
Can any dev try this and confirm if its a bug ?
I cant believe i lost all my courses and there is no backup folder in Courseplay mod settings folder in My Documents.....:/
Btw, i use latest version of CP.

Tensuko:
Found another problem, but that is not how you did that.
To do ANY action inside the manager it is always:
Click the action you want to do, THEN select the course.

So for move it is:
- Click on move entry
- Click on the course
- Click on the new folder

However, when select "move entry" and do a double click on a folder, it will be deleted, that might happened to you then.
@schwiti6190 easy to reproduce, guess need a savety check or something there.

PR #2048, opened 2022-10-28
Fix for #2045

schwiti6190:


Tensuko:
Move entry -> double click on the folder delets the folder.
But might found out the problem.
When I click move entry, click one folder, then another, it moves the folder into another folder.
I guess due to folder can't be moved into itself, it gets deleted.

fsScreen_2022_10_28_21_47_26

And it also moved the courses of the folder (somehow like you tried that in the 1st version of the manager).

PR #2102, opened 2022-11-15
Bunker silo driver

Tensuko:
- works with a shield (direction to drive into bunker, is set automatically)
- works with a roller (direction can be set, since the direction doesn't matter)
- works with a spreader (e.g. Fliegel Duo)
- the Bunker must be selected on the map with the target icon, the Giants AI Marker also sets the park position (Giants Helper can, but must not be activated)
- the rest is done with the pathfinder
- when a vehicle with a full trailer approaches, the driver will automatically drive to the park position


scripts/ai/AIUtil.lua
```diff
@@ -331,6 +331,12 @@ function AIUtil.getFirstAttachedImplement(vehicle,suppressLog)
 			if not suppressLog then
 				CpUtil.debugVehicle(CpDebug.DBG_IMPLEMENTS, vehicle, '%s front distance %d', implement.object:getName(), d)
 			end
+			if implement.object.spec_leveler then
```
pvaiko, line 334: Can we move this up over the log so it appears in the log?

scripts/ai/AIUtil.lua
```diff
@@ -354,6 +360,12 @@ function AIUtil.getLastAttachedImplement(vehicle,suppressLog)
 			if not suppressLog then
 				CpUtil.debugVehicle(CpDebug.DBG_IMPLEMENTS, vehicle, '%s back distance %d', implement.object:getName(), d)
 			end
+			if implement.object.spec_leveler then
```
pvaiko, line 363: Same here

scripts/ai/ProximityController.lua
```diff
@@ -65,6 +65,20 @@ function ProximityController:onBlockingVehicle(vehicle, isBack)
     end
 end
 
+--- Registers a function to ignore a object or vehicle.
+function ProximityController:registerIgnoreObjectCallback(object, callback)
```
pvaiko, line 69: This does not appear to be used anywhere. Also, the sensor can already ignore vehicles, may consider using that mechanism.

pvaiko (commented):
I only looked at the changes of the exisiting code, have a few minor comments.

scripts/ai/ProximityController.lua
schwiti6190: @pvaiko It's currently used to ignore the bunker silo walls for the proximitiy controller in the new strategy.
I will leave a TODO comment above the function and we consolidate this with vehicle at a later time?

pvaiko (approved):

PR #2104, opened 2022-11-15
Remaining time

Tensuko:
The remaining time is just a guess about some fixed values and just for fieldwork.
Refilling, standing arround (e.g. harvester waiting for an unloader) is ofc not part of the calculation.

### DIFF
--- a/scripts/gui/CpStatus.lua
+++ b/scripts/gui/CpStatus.lua
@@ -22,12 +22,18 @@ function CpStatus:reset()
     self.remainingTime:reset()
 end
 
+function CpStatus:start()
+    self.remainingTime:start()
+end
+
 function CpStatus:setActive(active)
     if self.isActive ~= active then 
         self.isActive = active
         self:raiseDirtyFlag()
         if not active then 
             self:reset()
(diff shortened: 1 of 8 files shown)


scripts/gui/CpStatus.lua
```diff
@@ -5,37 +5,56 @@ CpStatus = CpObject()
...
+    end 
+    self:setActive(isActive)
+end
+
+function CpStatus:setWaypointData(currentWaypointIx, numberOfWaypoints, course)
     if self.currentWaypointIx ~= currentWaypointIx then 
         self.currentWaypointIx = currentWaypointIx
         self.numberOfWaypoints = numberOfWaypoints
         self:raiseDirtyFlag()
+        self.remainingTime:calculate(course, currentWaypointIx)
```
pvaiko, line 56: Why is the calculation (and the remainingTime object) in the HUD, why not in the fieldwork strategy, should the HUD really have a dependency on the course, and not just a stupid display of values provided by others?

pvaiko (changes requested):

PR #2110, opened 2022-11-16
New Visibility Mode: Show Waypoints around current Waypoint.

tn4799:
This new Visibility Mode for the course allows the user to see directly where the tractor is heading next or was about to head next. This way he can track the work on the field without having the clutter of seeing all waypoints.


scripts/gui/CourseDisplay.lua
```diff
@@ -259,22 +259,36 @@ function SimpleCourseDisplay:updateChangesBetween(firstIx, secondIx)
 end
 
 --- Changes the visibility of the course.
-function SimpleCourseDisplay:updateVisibility(visible, onlyStartStopVisible)
+function SimpleCourseDisplay:updateVisibility(visibilityData)
```
pvaiko, line 262: I would prefer keeping the original boolean parameters (maybe rename visible to allVisible) and adding a onlyCurrentVisible (not last, as that may also mean the last waypoint of the course) and a currentWaypointIx or so parameter.

scripts/gui/CourseDisplay.lua
```diff
@@ -259,22 +259,36 @@ function SimpleCourseDisplay:updateChangesBetween(firstIx, secondIx)
...
-function SimpleCourseDisplay:updateVisibility(visible, onlyStartStopVisible)
+function SimpleCourseDisplay:updateVisibility(visibilityData)
 	if self.course then
 		local numWp = self.course:getNumberOfWaypoints()
 		for j = 1, numWp do
-			if self.signs[j] then 
+			if self.signs[j] then
+				local visible = false
+				local onlyStartStopVisible = false
+				if visibilityData.visiblityMode == CpVehicleSettings.SHOW_COURSE_ALL then
```
pvaiko, line 269: visiblityMode is mistyped.

config/CombineUnloaderJobParameterSetup.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pvaiko, line 1: Why is this file in the pull request?

config/VehicleConfigurations.xml
```diff
@@ -333,6 +333,13 @@ You can define the following custom settings:
     />
```
pvaiko, line 333: Why is this file in the pull request?

scripts/gui/CourseDisplay.lua
```diff
@@ -259,22 +259,36 @@ function SimpleCourseDisplay:updateChangesBetween(firstIx, secondIx)
...
+					visible = visibilityData.isControlled ~= nil and visibilityData.isControlled or visibilityData.isControlled == nil
```
pvaiko, line 270: just `visible = visibilityData.isControlled` would work fine, no need to check for nil if we use it as boolean.

But again, this logic should be in the caller, as it was before.

scripts/specializations/CpCourseManager.lua
```diff
@@ -459,6 +463,13 @@ end
 function CpCourseManager:onUpdate()
     local spec = self.spec_cpCourseManager
     spec.courseRecorder:update()
+
+    --update lastWaypoint variable to keep moving the course when passing a new waypoint.
+    --only needs to be done when in show current waypoints mode to save performance
```
pvaiko, line 468: Even better, don't call this on every update cycle as the current waypoint changes only few seconds.
Call for instance from `AIDriveStrategyFieldWorkCourse:onWaypointPassed`

pvaiko (changes requested):
Good idea.

scripts/gui/CourseDisplay.lua
tn4799: The reason why ive done this is to make it easier to maintain and also add new visibility modes.
This is possible due to having all logic in one place instead of at every function call. 
Also with the packed data inside the table it is easy to add new parameters without having to adjust the whole declaration  of the function.

scripts/specializations/CpCourseManager.lua
tn4799: Thats smart. I havnt found that. But i will change it.

scripts/gui/CourseDisplay.lua
pvaiko: The idea here is that the SimpleCourseDisplay class only cares about displaying the waypoints and nothing else (loose coupling), and so, does not have any dependency on some settings or, for that matter if the vehicle is controlled or not. It is dumb, we have to tell it only which waypoints (if any) to show. 

So yes, I'd still prefer keeping the decision which waypoints to show in the caller, and pass on very simple instructions to this class only.

tn4799: So if I write one centralized function in the course manager that contains all the logic and then passes the logical values through to the simpleCourseDisplay this would be fine, am I right? 

So we have a centralized logic to make maintainance and maybe new visibilityModes easier to add, but still keep the loose coupling, which is in my eyes the best of both worlds.

tn4799:
General Question: Is it ok if i leave the trim trailing whitespaces setting on? otherwise i always have to remember to turn the setting off when i write some code for courseplay.

scripts/gui/CourseDisplay.lua
pvaiko: That sounds ok to me, thanks.

pvaiko:
> 

Sorry, I did not see this, yes, it is ok.

schwiti6190:
Closes as a new PR was created #2140

PR #2234, opened 2023-01-05
Beds from ouside in?

Silberfussel:
Hi
There is the option to process beds from the inside out, but what about the other way around? So work off the beds from the outside in. Would it be possible?


Tensuko:
Do not open such questions as a PR pls.
And yes that is start on center and not headland option on course generator.

PR #2242, opened 2023-01-09
Update PL translate (v7.2.0.0)

KITT3000:


Tensuko:
Sry buddy, for ModHub a bit too late, it is already in testing... Giants was super fast this time....
But I am sure that the next ModHub Update isn't too far away,

KITT3000:
np ;) My fault because I haven't checked github in a long time

Tensuko:
Had to fix something, so re sent and translation is now in :)

PR #2243, opened 2023-01-09
Update translation_fr.xml

Lactic68:
v7.2.0.0


Tensuko:
Also too late for modhub. (so far)
This was already part of 7.1.2.5 but I guess we won't have too long for a next modhub Update.

Giants was super fast this time with in testing state....

Lactic68:
Yep, sorry for the translation delay. Anyway always congrats for all your hard work to make this mod better each time and happy new year to the team.

Tensuko:
Happy new year for you too :) 
As we got cp back I could put in the translation too so its fine :D

PR #2276, opened 2023-01-27
Issue 2265 and small bug fix 

schwiti6190:
Test: 
- Gui with this vehicle: https://www.farming-simulator.com/mod.php?lang=de&country=de&mod_id=262659&title=fs2022
- Testing of bunker silo driver to make sure nothing broke there.
- Test field border drawing and if the disabled hotspots are correct.


5g4M:
The Bunker fix helped , but some tractors are still behaving weird. I would say atleast 65% of those that I tested seem to work proberly so far.

PR #2420, opened 2023-03-29
Added leftover bales to cp status and hud page for balefinder

schwiti6190:


translations/translation_br.xml
```diff
@@ -63,6 +63,7 @@
     <text name="CP_baleFinderJobParameters_baleWrapType_all_bales" text="todos os fardos"/>
     <text name="CP_baleFinderJobParameters_baleWrapType_only_wrapped_bales" text="somente embalado"/>
     <text name="CP_baleFinderJobParameters_baleWrapType_only_not_wrapped_bales" text="não embalado"/>
+    <text name="CP_baleFinder_balesLeftover" text="Bales leftover: "/>
```
pvaiko, line 66: "Bales left:" sounds here better than leftover. Also in symbol names.

scripts/ai/AIDriveStrategyFindBales.lua
```diff
@@ -483,13 +488,15 @@ function AIDriveStrategyFindBales:approachBale()
         if not self:isReadyToLoadNextBale() then
             self:debug('Start picking up bale')
             self.state = self.states.WORKING_ON_BALE
+            self:findBales() --- Refreshes for bale counter
```
pvaiko, line 491: Just a bit worried about the performance here with big fields and many bales, one could just decrement self.balesLeft until we'd look for new bales anyway, but may not be a problem at all.

pvaiko (changes requested):

Issue #2449, opened 2023-04-12
[BUG_MP] Dedicated Server Sync stopping at 99%


PapaTim68:
After the update to the latest Version 7.2.1.0 joining the dedicated Server doesn't work and is stuck at the Loading Screen.
Single Player seems to load fine with that Version.

**Courseplay version**
Version used: 7.2.1.0

**Client Log**
Line 4344
```
2023-04-12 11:36 Error: Running LUA method 'packetReceived'.
C:/Users/Tim/Documents/My Games/FarmingSimulator2022/mods/FS22_Courseplay/scripts/ai/jobs/CpJobParameters.lua:240: attempt to index local 'dischargeNode' (a nil value)
```
[log_client.txt](https://github.com/Courseplay/Courseplay_FS22/files/11209884/log_client.txt)

**Server Log**
Server log seems to not show anything.
[log_server.txt](https://github.com/Courseplay/Courseplay_FS22/files/11209887/log_server.txt)

**Player on the server**
0 Players directly after Restart due to update of Courseplay

**Dedicated server or local hosted server**
Dedicated server

**Map**
NORTH FRISIAN MARCH 4X oG

PapaTim68:
Moin,
tried out the new Version, but sadly still has the same Error.
If needed, new Logs with this version:
[log_client.txt](https://github.com/Courseplay/Courseplay_FS22/files/11239002/log_client.txt)


[log_server.txt](https://github.com/Courseplay/Courseplay_FS22/files/11239003/log_server.txt)

As far as I can tell, it's the same Error.
But I noticed that Headland Management has a line about the Initialization of the Holmer Mod directly after the error. Since the error is related to Discharge Nodes, the error might be related to Holmer's Quick exchange transfer wagon.

PapaTim68:
Can confirm it's loading now!

Danke :-)

Issue #2443, opened 2023-04-08
[BUG_SP] Combine folds auger in extended position when CP is started with extended auger


DemsMani:
**READ FIRST**
https://github.com/Courseplay/Courseplay_FS22/wiki

**Steps to reproduce**
When the auger of the combine is extended and cp is started again (for example after a collision), the outer link of the auger folds in, leaving no option anymore to do anything

**Actual behavior**
When pressing the cp start button with the auger extended, the outer link just folds in immediately

**Expected behavior**
As it worked before, the outer link of the auger should stay unfolded

**Screenshots**
![Farming Simulator 22 08_04_2023 23_23_01](https://user-images.githubusercontent.com/99053368/230743743-5185e737-318c-4a24-8963-860f166af2b0.jpg)

**Courseplay version**
Version used: 7.2.1.0

**Log**
(https://gist.github.com/DemsMani/a7d0fdc415abc2c0b7c816cf5a865658)

**Map**
Morris

**Vehicles/ implements involved**
Lexion 8900 (Modhub)

**Relevant script mods**
Autodrive (no active driver during test)

yann-mrt:
It is not isolated case, this happen to me as well with V 7.2.1.0, everything was find with V 7.2.0.5.

Tensuko:
No one said anything different...

knoxville73:
Hi,

I know the same issue with an another combine

Issue #2440, opened 2023-04-08
Problem with pipe offset when offloading combine[BUG_SP]


Martin-L45:
**READ FIRST**
https://github.com/Courseplay/Courseplay_FS22/wiki

**Steps to reproduce**
Steps to reproduce the behavior.
Use courseplay to unload combine

**Actual behavior**
A clear and concise description of what happen.
In the V7.2.1.0 version the offloader does not drive in the correct position. I need to set 4m offset and it offloads in the front of the augerwagon
This is not a problem in V7.2.0.5

**Expected behavior**
Drive in the correct position

**Screenshots**
V7.2.1.0
![Farming Simulator 22 Screenshot 2023 04 08 - 12 37 35 80](https://user-images.githubusercontent.com/85753728/230716952-5cf8d6aa-b7b4-4299-9fb0-0beba384beb0.png)
V7.2.0.5
![Farming Simulator 22 Screenshot 2023 04 08 - 12 34 26 79](https://user-images.githubusercontent.com/85753728/230716955-198d31c4-3297-4cc7-a171-d58841102d0a.png)

**Courseplay version**
Version used: 

**Log**
Please attach a log with all debug channels active.
For help with the debug channels,
please check out the readme at the  github front page.

[logV7210.txt](https://github.com/Courseplay/Courseplay_FS22/files/11183021/logV7210.txt)
[logV7205.txt](https://github.com/Courseplay/Courseplay_FS22/files/11183022/logV7205.txt)

**Map**
Map played on.

**Vehicles/ implements involved**
Which vehicle or implement?
Mod yes/no ?

**Relevant script mods**

Mods like Autodrive or VCA and so on.

Tensuko:
At the front is normal and correct, the unloader will drive further to the depending on the fillvolume of the trailer. Wonder if we have this missing for the auger wagon. Will check on that.

PR #2452, opened 2023-04-14
Fix for #2444, #2449 and possible improvement for #2443 , #2440

schwiti6190:


scripts/ai/controllers/PipeController.lua
```diff
@@ -285,12 +286,27 @@ function PipeController:measurePipeProperties()
...
-        self.implement:setAnimationTime(self.pipeSpec.animation.name, 1, true, false)
+function PipeController:instantUnfold(ignorePipeState)
+    if self.pipeSpec.currentState ~= self.pipeSpec.targetState or 
+        self.pipeSpec.currentState ~= PipeController.PIPE_STATE_OPEN or ignorePipeState then
+        if self.foldableSpec.turnOnFoldDirection == -1 then
+            Foldable.setAnimTime(self.implement, 0, true)
+        else 
+            Foldable.setAnimTime(self.implement, 1, true)
+        end
+        if self.pipeSpec.hasMovablePipe then
```
pvaiko, line 297: We should really explain why are we doing this, open/close/open, plus some animation magic is far from self explanatory, will help us later to remember what this hack is doing.

pvaiko (changes requested):

scripts/ai/controllers/PipeController.lua
schwiti6190: Of cause you are right, but I don't really now how I made this working in first place. Mostly alot of try and error and praying.

PR #2609, opened 2023-07-06
Enables unloading of the cutter header from an attached trailer at th…

schwiti6190:
…e field.

Combine picks up the header of the trailer, which needs to be attached.
1) When the harvester is started from the hud, then the current position is used to store the trailer.
2) Starting from the AI menu uses the giants marker to store the trailer.


scripts/ai/AIDriveStrategyAttachHeader.lua
```diff
@@ -0,0 +1,306 @@
+--[[
+This file is part of Courseplay (https://github.com/Courseplay/Courseplay_FS22)
+Copyright (C) 2022 Courseplay Dev Team
```
pvaiko, line 3: Year

scripts/ai/AIDriveStrategyAttachHeader.lua
```diff
@@ -0,0 +1,306 @@
...
+    else 
+        self.mode = self.MODES.ATTACH_HEADER_FROM_ATTACHED_TRAILER
+        self:debug("A cutter/header on a separate trailer is attached on the back.")
+        self.trailer = AIUtil.getImplementWithSpecialization(self.vehicle, DynamicMountAttacher)
+        self.dynamicMountAttacherController = DynamicMountAttacherController(vehicle, self.trailer)
+    end
+    self.attachableController = AttachableController(vehicle, self.trailer)
+    self.attacherJointController = AttacherJointController(vehicle, vehicle)
+    self:appendImplementController(self.attacherJointController)
+    local trailerAreaNode = createTransformGroup("tempTrailerNode")
```
pvaiko, line 89: Does not this needs to be removed after we don't need it?

scripts/ai/AIDriveStrategyAttachHeader.lua
```diff
@@ -0,0 +1,306 @@
...
+    self.trailerAreaToAvoid =  PathfinderUtil.NodeArea(trailerAreaNode, -self.trailer.size.width/2, -self.trailer.size.length/2, self.trailer.size.width, self.trailer.size.length)
```
pvaiko, line 90: I might be wrong, and anyway, may not be a problem but this assumes that the cutter node is at the root node of the trailer and the root node is exactly in the middle of trailer (they sometimes have a lengthOffset for the size)

scripts/ai/AIDriveStrategyAttachHeader.lua
```diff
@@ -0,0 +1,306 @@
...
+            goalNode = self.dynamicMountAttacherController:getCutterJointPositionNode()
+            rootNode = self.dynamicMountAttacherController:getMountedImplement().rootNode
+        else 
+            goalNode = self.attachableController:getCutterJointPositionNode()
+            rootNode = self.trailer.rootNode
+        end
+        local x, _, z = getWorldTranslation(goalNode)
+        local dirX, _, dirZ = localDirectionToWorld(rootNode, 0, 0, 1)
+        local rotY = MathUtil.getYRotationFromDirection(dirX, dirZ)
+        self.cutterNode = CpUtil.createNode("cutterNode", x, z, rotY)
```
pvaiko, line 235: Shouldn't this only create it if does not yet exist?

scripts/ai/controllers/AttachableController.lua
```diff
@@ -0,0 +1,60 @@
+--- This controller is for implement, that can be attach to vehicles.
```
pvaiko, line 1: A spell check for these few lines maybe :)

scripts/ai/AIDriveStrategyAttachHeader.lua
schwiti6190: A good point I am not even sure if this check is even needed. I guess I will put a comment there and we will see if it breaks from a user test.

pvaiko: Yeah, you right, the pathfinder should detect a collision there and avoid it anyway.

Issue #2677, opened 2023-08-02
[BUG_SP] Low Loaders unfold when Bale collecting using Universal Autoload


armyf35:
**Steps to reproduce**
1. Create some bales on a field
2. Get a low loader that can unfold (DTS 5.9 | Steel Drop Deck | Renown Drop Deck)
3. Get a truck
4. Attach to low loader
5. Target field with the bales created in step 1
6. Drive near/into the field
7. Start Courseplay driver

**Actual behavior**
Driver first unfolds the trailer before attempting to pickup bales. Since these are low loaders (And using Universal Autoload to load bales) not bale loaders, there is no physical part needed to pickup bales. In the case of the DTS 5.9 specifically, legs are dropped at the rear of the trailer which causes it to slide around since the rear wheels are off the ground.

**Expected behavior**
The driver should not unfold low loader trailers when bale loading since this can cause steering issues and/or needlessly increase the length of the trailer.

**Screenshots**
![1](https://github.com/Courseplay/Courseplay_FS22/assets/7518618/8ca3a6dc-93f3-4f25-a063-7a016a8da041)
![2](https://github.com/Courseplay/Courseplay_FS22/assets/7518618/df2d80e7-cc9b-4d4e-8b51-a5f6b0c2a4a9)
![3](https://github.com/Courseplay/Courseplay_FS22/assets/7518618/9f8bb9f0-9fe5-4a4e-be92-79824dd3db34)
![4](https://github.com/Courseplay/Courseplay_FS22/assets/7518618/90c38cdd-3dcc-4ff0-8345-31b38d571e91)
![5](https://github.com/Courseplay/Courseplay_FS22/assets/7518618/b18b3836-71cf-424d-b75d-550e0d598103)

**Courseplay version**
Version used: 7.2.1.9

**Log**
[log.txt](https://gist.github.com/armyf35/059d9c0e7dc51356138fb30d9954097c)

**Map**
Tested on Elmcreek

**Vehicles/ implements involved**
- Low Loaders (Tested with all four base game loaders, the "Transportwagen TP 500S" cannot fold and therefore has no issue)
  - DTS 5.9
  - Steel Drop Deck
  - Renown Drop Deck
  - Low Loader 4A (Platinum Expansion)
  - Low Loader 3A (Platinum Expansion)

**Relevant script mods**
- Universal Autoload (1.4.1.0)
- Easy Development Controls (1.3.0.2)

Tensuko:
[...]

PR #2678, opened 2023-08-02
Vehicle config to avoid Issue 2677

schwiti6190:
- Fix for #2677 
- Added new vehicle config: "disableUnfoldBaleLoader" to disable bale loader unfold.
- New console commands for working with vehicle configurations:
  - Print entered vehicle and implement config xml filenames.
  - Print attribute for entered vehicle and implements.
  - Print all assigned attributes for entered vehicle and implements.
  - Print all valid attribute names
- All config commands are now under: "cpVehicleConfigurations..."


scripts/ai/controllers/FoldableController.lua
```diff
@@ -0,0 +1,42 @@
+--[[
+This file is part of Courseplay (https://github.com/Courseplay/courseplay)
+Copyright (C) 2019-2022 Peter Vaiko
```
pvaiko, line 3: year, author

scripts/specializations/CpAIWorker.lua
```diff
@@ -504,4 +505,33 @@ function CpAIWorker:onStartAutoDrive()
...
+    g_devHelper.consoleCommands:registerConsoleCommand("cpVehicleOnWorkStartTest", 
+        "Raise work start command", "consoleCommandRaiseWorkStart", CpAIWorker)
+    --- TODO: Add function to simulate lowering and raising with controllers.
+end
+
+--- Helper command to test the pipe measurement.
+function CpAIBaleFinder:consoleCommandRaiseWorkStart()
+    local vehicle = g_currentMission.controlledVehicle
+    if not vehicle then 
+        CpUtil.info("Could not measure pipe properties without entering a vehicle!")
```
pvaiko, line 524: This probably isn't the correct log message.

scripts/specializations/CpAIWorker.lua
```diff
@@ -504,4 +505,33 @@ function CpAIWorker:onStartAutoDrive()
...
+--- Console commands
+---------------------------------------------
+
+function CpAIWorker.registerConsoleCommands()
+    g_devHelper.consoleCommands:registerConsoleCommand("cpVehicleOnWorkStartTest", 
+        "Raise work start command", "consoleCommandRaiseWorkStart", CpAIWorker)
+    --- TODO: Add function to simulate lowering and raising with controllers.
+end
+
+--- Helper command to test the pipe measurement.
```
pvaiko, line 520: pipe?

scripts/specializations/CpAIWorker.lua
```diff
@@ -504,4 +505,33 @@ function CpAIWorker:onStartAutoDrive()
...
\ No newline at end of file
+end
+
+---------------------------------------------
+--- Console commands
+---------------------------------------------
+
+function CpAIWorker.registerConsoleCommands()
+    g_devHelper.consoleCommands:registerConsoleCommand("cpVehicleOnWorkStartTest", 
+        "Raise work start command", "consoleCommandRaiseWorkStart", CpAIWorker)
```
pvaiko, line 516: It is a bit confusing what this is doing especially it seems to raise an event (not raise the implements) which lowers the implements. Also, only foldable controllers are notified, so its scope seems to be the foldables only, what about the other controllers?

pvaiko (changes requested):

scripts/specializations/CpAIWorker.lua
schwiti6190: Forgot to comment my intentions there, was to acting to fast ...
For now I only needed the foldable controller for testing and the other could be added later when needed.

Tensuko:
@schwiti6190 conflict here now too, also what is the status?

Tensuko:
Rebased and added config and tested it with AutoLoad from Loki, does work.
Anything missing?

Issue #2681, opened 2023-08-03
[BUG_SP] Holmer Terra Variant 435 with Joker 8 RT bug with the sowing and fertilizing option.


schwiti6190:
**Steps to reproduce**
Holmer Terra Variant 435 with Joker 8 RT bug with the sowing and fertilizing option.

PR #2682, opened 2023-08-03
Might be a fix for #2681

schwiti6190:


Tensuko:
When both options (seeding and fertilizing) are active , every 2nd time I start the driver the msg "needs to be filled" pops up, hit start again and it works fine.
[log.txt](https://github.com/Courseplay/Courseplay_FS22/files/12261916/log.txt)

Tensuko (approved):

PR #2711, opened 2023-08-17
Bug Fix Fixes #2710

### LINKED ISSUE #2710: [BUG_SP] Inital State in Combine Unloader creates a no path found condition
When resuming a combine unloader on field unload that is above the trailer fill level it will create a no path found condition. The idle state is set first time this code is ran. Then the idle state will be set again while waiting on pathfinder. Logic downstream that checks to see if a path is found includes a state check which fails because it is looking for the waiting for pathfinder state. I put debug prompts in the code to determine this and it printed three times. Log https://pastebin.com/HnXGQ6dg. ```self:debug('I am setting idle")``` was placed before line 390

https://github.com/Courseplay/Courseplay_FS22/blob/b2574c51a371eafa630f8afe26df45172a60eecf/scripts/ai/AIDriveStrategyUnloadCombine.lua#L385-L391

pops64:
Doesn't prevent creating multiple timers in the 50ms delay in the first one to get out of the initial state but will prevent a state change once beyond initial

### DIFF
--- a/scripts/ai/AIDriveStrategyUnloadCombine.lua
+++ b/scripts/ai/AIDriveStrategyUnloadCombine.lua
@@ -384,10 +384,12 @@ function AIDriveStrategyUnloadCombine:getDriveData(dt, vX, vY, vZ)
         end
     elseif self.state == self.states.INITIAL then
         Timer.createOneshot(50, function ()
-            --- Pipe measurement seems to be buggy with a few over loaders, like bergman RRW 500,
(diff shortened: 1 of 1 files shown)


pvaiko (approved):

pops64:
Add the fix suggest by @schwiti6190 in the issue related to this

Tensuko (approved):

schwiti6190 (approved):

Issue #2701, opened 2023-08-09
[Feature] Add Support for Choppers/Sugcane back in


pops64:
CP as always supported choppers since at least FS15(When I started playing). I would love to see them brought back in with support. Most of the code to support this is already there just needs a few tweaks to handle some of the particularities of  Forage Harvester Unloading.

Looked through the code base and from experience in 17. The missing items to add this functionality back in are as follows.

- [x] Combine Driver: Needs to wait until tipper is in range unless turning(Code not present) GDNs [website](https://gdn.giants-software.com/documentation_scripting_fs22.php?version=script&category=1&class=4) on how they do this
- [x] Combine Unloading Driver Needs to handle end of row turn around. Moving out of the way and looping around with enough space to realign with Chopper. The tipper needs be checked if reversible this has and can cause issues with turn maneuvers. It appears most of the code needed to achieve this is already present in CP just need to combine the parts together 
- [x] Combine Unloading Driver needs to handle headlands or disable it complete. Any sharp curve in the headland maneuver caused issues in FS17. The only time CP FS17 was able to handle headlands with choppers where when the course gen was set to Sharp forcing the driver to handle it like a turn. Otherwise it wouldn't be able to stay in range of the chopper or it would collide depending on the curve
- [x] Shifting the Combine Unloading Driver call to chopper when current tipper is almost full instead of the combine as currently coded(Not sure if this is coded when multiple tippers are unloading a combine)
- [x] Choppers can unload both left and right side. An option to pick which side or just always pick the side without fruit(No option always non fruit side 95% of the time)

I really hope this can be added back in. As babysitting a chopper is a PITA and time can be better spent else where. I see autodrive can handle? Never really used AD it was always clunky and I always enjoyed CP being the one stop shop of FS helper automation

Tensuko:
Nothing will be imported from FS17.
If this will be put back into FS22, will import it from FS19 but it still needs more adjustments and needs to be put into a chopperUnloaderDriver to match our new file organisation and not end up as chaotic as in FS19 and before.

Chopper is waiting untill can unload, not sure if there is a big difference, might can adjust this.

The rest will come from FS19 + adjustments that are needed.

However it was never planed so can't rly say if we put it back in.
Maybe next year? Who knows....

There is other stuff that needs to be done first.

pops64:
[...]

PR #2712, opened 2023-08-17
Chopper support

pops64:
Currently 1 Tipper is supported. Correctly selects non fruit side and turns when chopper is finished. Looking for testers as there are lots of edges cases need to be done. 
Things to test as this is almost complete

- [ ] Combines - Nothing Changed code was moved around
- [x] Headlands - Sharp turns or tight turns seems to handle rounded turns just fine
- [x] Sugarcane - Should work but I haven't coded or tested anything. Small work width might cause issues
- [x] AD Interactions Both Chaff and Sugarcane
- [ ] Grass - Haven't tested, this might perform poorly due grass not being a fruit
- [x] Calling a 2nd tipper when current is almost full. This might require some effort but I would like to get it done before finalizing

### DIFF
--- a/config/jobParameters/CombineUnloaderJobParameterSetup.xml
+++ b/config/jobParameters/CombineUnloaderJobParameterSetup.xml
@@ -17,10 +17,12 @@
 			<Values>
 				<Value name="UNLOAD_COMBINE">1</Value>
 				<Value name="UNLOAD_SILO_LOADER">2</Value>
+				<Value name="UNLOAD_CHOPPER">3</Value>
 			</Values>
 			<Texts>
 				<Text>combine</Text>
 				<Text>siloLoader</Text>
+				<Text>chopper</Text>
 			</Texts>
 		</Setting>
 	</SettingSubTitle>
--- a/scripts/ai/AIDriveStrategyChopperCourse.lua
+++ b/scripts/ai/AIDriveStrategyChopperCourse.lua
@@ -389,11 +389,6 @@ function AIDriveStrategyChopperCourse:getUnloader(driver)
     elseif driver == self:getNextUnloader() then
         return self:getNextUnloader()
     end
-    -- if whichUnloader == 'A' then
-    --     return self.unloaders.unloaderA:get()
-    -- elseif whichUnloader == 'B' then
-    --     return self.unloaders.unloaderB:get()
-    -- end
 end
 
 function AIDriveStrategyChopperCourse:updateNextUnloader()
@@ -405,22 +400,7 @@ function AIDriveStrategyChopperCourse:updateNextUnloader()
         self:resetUnloader(self:getNextUnloader())
     end
 
-    -- if self.unloaders.currentUnloader == 'A' then
-    --     self.unloaders.nextUnloader = 'A'
-    --     self.unloaders.currentUnloader = 'B'
-    -- else
-    --     self.unloaders.nextUnloader = 'B'
-    --     self.unloaders.currentUnloader = 'A'
-    -- end
 end
-
--- function AIDriveStrategyChopperCourse:getNextUnloaderID()
---     return self.unloaders.nextUnloader
--- end
-
--- function AIDriveStrategyChopperCourse:getCurrentUnloaderID()
---     return self.unloaders.currentUnloader
--- end
 
 function AIDriveStrategyChopperCourse:getCurrentUnloader()
     return self.unloaders.currentUnloader:get()
--- a/scripts/ai/AIDriveStrategyChopperCourse.lua
+++ b/scripts/ai/AIDriveStrategyChopperCourse.lua
@@ -108,6 +108,12 @@ function AIDriveStrategyChopperCourse:getDriveData(dt, vX, vY, vZ)
     return AIDriveStrategyFieldWorkCourse.getDriveData(self, dt, vX, vY, vZ)
 end
 
+function AIDriveStrategyChopperCourse:start(course, startIx, jobParameters)
+    AIDriveStrategyChopperCourse.superClass().start(self, course, startIx, jobParameters)
+    -- Update the pipeOffset side when we start work  
+    self:updatePipeOffset(startIx)
+end
+
 function AIDriveStrategyCombineCourse:driveUnloadOnField()
     if self.unloadState == self.states.STOPPING_FOR_UNLOAD then
         self:setMaxSpeed(0)
@@ -136,7 +142,7 @@ function AIDriveStrategyChopperCourse:estimateDistanceUntilFull(ix)
     local fillLevel = 1
     local capacity = 1
 
-    -- Choppers don't have fill levels get the trailer we currently are discharging too fill levels
+    -- Choppers don't have fill levels get the trailer we currently are discharging too fill levels. This is for when we have multiple tippers
     fillLevel, capacity = self:getTrailerFillLevel()
 
     if ix > 1 then
@@ -228,36 +234,6 @@ function AIDriveStrategyChopperCourse:getTrailerFillLevel()
     return fillLevel, capacity
 end
 
-function AIDriveStrategyChopperCourse:getFillLevelPercentage()
-    -- Hack to get the combineUnloader driver working
-    return 1
-end
-
---- Make life easier for unloaders, increases reach of the pipe
---- Old code ?? Not used. TO BE DELETED
-function AIDriveStrategyChopperCourse:fixMaxRotationLimit()
-    if self.pipe then
-        local lastPipeNode = self.pipe.nodes and self.pipe.nodes[#self.pipe.nodes]
-        if self:isChopper() and lastPipeNode and lastPipeNode.maxRotationLimits then
-            self.oldLastPipeNodeMaxRotationLimit = lastPipeNode.maxRotationLimits
-            self:debug('Chopper fix maxRotationLimits, old Values: x=%s, y= %s, z =%s', tostring(lastPipeNode.maxRotationLimits[1]), tostring(lastPipeNode.maxRotationLimits[2]), tostring(lastPipeNode.maxRotationLimits[3]))
-            lastPipeNode.maxRotationLimits = nil
-        end
-    end
-end
-
---- Old code ?? Not used. TO BE DELETED
-function AIDriveStrategyChopperCourse:resetFixMaxRotationLimit()
-    if self.pipe then
-        local lastPipeNode = self.pipe.nodes and self.pipe.nodes[#self.pipe.nodes]
-        if lastPipeNode and self.oldLastPipeNodeMaxRotationLimit then
-            lastPipeNode.maxRotationLimits = self.oldLastPipeNodeMaxRotationLimit
-            self:debug('Chopper: reset maxRotationLimits is x=%s, y= %s, z =%s', tostring(lastPipeNode.maxRotationLimits[1]), tostring(lastPipeNode.maxRotationLimits[3]), tostring(lastPipeNode.maxRotationLimits[3]))
-            self.oldLastPipeNodeMaxRotationLimit = nil
-        end
-    end
-end
-
 function AIDriveStrategyChopperCourse:isChopperWaitingForUnloader()
     return self.waitingForTrailer
 end
@@ -344,14 +320,6 @@ function AIDriveStrategyChopperCourse:checkFruit()
             self.fruitLeft, self.fruitRight, tostring(self.fieldOnLeft), tostring(self.fieldOnRight))
 end
 
--- even if there is a trailer in range, we should not start moving until the pipe is turned towards the
--- trailer and can start discharging. This returning true does not mean there's a trailer under the pipe,
--- this seems more like for choppers to check if there's a potential target around
--- Not being used. Replaced by other functions TO BE DELETED
-function AIDriveStrategyChopperCourse:canDischarge()
-    return self.pipeController:getDischargeObject()
-end
-
 -- Return true for no headlands so we can reduce off field pently for our unloader driver
 function AIDriveStrategyChopperCourse:hasNoHeadlands()
     return self.course:getNumberOfHeadlands() == 0
@@ -404,18 +372,3 @@ function AIDriveStrategyChopperCourse:updatePipeOffset(ix)
     --TODO Add a check to see if we are on the last row and if we are set the it to field side
     self:debug('I have left the pipeoffset on the right side')
 end
-
-function AIDriveStrategyChopperCourse:start(course, startIx, jobParameters)
-    AIDriveStrategyChopperCourse.superClass().start(self, course, startIx, jobParameters)
-    -- Update the pipeOffset side when we start work  
-    self:updatePipeOffset(startIx)
-end
-
-
--- -- TODO this code snipet seems to be peppered through out differnt child classes this should be moved higher up the stack
--- function AIDriveStrategyChopperCourse:isFruitAtIx(ix, xOffset, zOffset)
---     local x + xOffset, _, z + zOffset = self.course:getWaypointPosition(ix)
---     local hasFruit = PathfinderUtil.hasFruit(x, z, 1, 1)
---     self:debug('isFruitAtIx %s, x = %.1f, z = %.1f (xOffset = %.1f, zOffset = %.1f', tostring(hasFruit), x, z, xOffset, zOffset)
---     return hasFruit
--- end
--- a/scripts/ai/AIDriveStrategyChopperCourse.lua
+++ b/scripts/ai/AIDriveStrategyChopperCourse.lua
@@ -137,7 +137,6 @@ function AIDriveStrategyChopperCourse:onWaypointPassed(ix, course)
     end
 
     self:checkFruit()
-    self:headingToUpDown(ix)
 
     -- make sure we start making a pocket while we still have some fill capacity left as we'll be
     -- harvesting fruit while making the pocket unless we have self unload turned on
@@ -170,11 +169,11 @@ function AIDriveStrategyChopperCourse:onWaypointPassed(ix, course)
     AIDriveStrategyFieldWorkCourse.onWaypointPassed(self, ix, course)
 end
 
-function AIDriveStrategyChopperCourse:start(course, startIx, jobParameters)
-    AIDriveStrategyChopperCourse.superClass().start(self, course, startIx, jobParameters)
-    -- Update the pipeOffset side when we start work  
-    self:updatePipeOffset(startIx)
-end
+-- function AIDriveStrategyChopperCourse:start(course, startIx, jobParameters)
+--     AIDriveStrategyChopperCourse.superClass().start(self, course, startIx, jobParameters)
+--     -- Update the pipeOffset side when we start work  
+--     self:updatePipeOffset(startIx)
+-- end
 
 function AIDriveStrategyChopperCourse:checkNextUnloader()
     if not self:getUnloader(self:getCurrentUnloaderIdent()) and self:getUnloader(self:getNextUnloader()) then
@@ -451,29 +450,34 @@ function AIDriveStrategyChopperCourse:updatePipeOffset(ix)
     -- Pipe in fruit map can't be used on headlands so always use hasFruit
     -- If fruit is found using our current pipe offset update to the opposite side
     
+    local fruitCheckWaypoint = ix or self.course:getCurrentWaypointIx()
+    self:debug('Fruitwaypoint was set to %d', fruitCheckWaypoint)
+
+    if not self.course:isOnHeadland(fruitCheckWaypoint) then
+        local lRow = self.course:getRowLength(fruitCheckWaypoint)
+        if ixAtRowStart then
+            fruitCheckWaypoint = self.course:getNextWaypointIxWithinDistance(fruitCheckWaypoint,lRow / 2)
+            self:debug('Fruitwaypoint was set to the middle of the row %d', fruitCheckWaypoint)
+        else
+            fruitCheckWaypoint = fruitCheckWaypoint + 20
+            self:debug('Fruitwaypoint was set 20 ahead to couldn\'t determine row length %d', fruitCheckWaypoint)
+        end
+    end
+
     -- Reset the pipeoffset if we where being chased by a unloader driver
     if self.pipeOffsetX == 0 then
         self:setPipeOffsetX()
         self.chaseMode = false
         self:setPipeOffsetZ()
+        self:debug('reset chase mode')
     end
 
-    local storedIx = ix
-    while ix <= storedIx + 10 do
-        if self.course:isTurnStartAtIx(ix) then
-            self:debug('There is a turn at %d check fruit at %d', ix, ix - 10)
-            ix = ix - 10
-            break
-        else
-            ix = ix + 1
-        end
-    end
-    local hasFruit = self:isPipeInFruitAtWaypointNow(self.course, ix, self.pipeOffsetX)
-    self:debug('I found fruit %s at waypoint %d', tostring(hasFruit), ix)
+    local hasFruit = self:isPipeInFruitAtWaypointNow(self.course, fruitCheckWaypoint, self.pipeOffsetX)
+    self:debug('I found fruit %s at waypoint %d', tostring(hasFruit), fruitCheckWaypoint)
     if hasFruit then
         self:debug('I found fruit use the opposite side')
         self.pipeOffsetX = -self.pipeOffsetX
-        hasFruit = self:isPipeInFruitAtWaypointNow(self.course, ix, self.pipeOffsetX)
+        hasFruit = self:isPipeInFruitAtWaypointNow(self.course, fruitCheckWaypoint, self.pipeOffsetX)
         if hasFruit then
             self:debug('I must be on a land row switch to chase mode')
             self.pipeOffsetX = 0
@@ -484,10 +488,10 @@ function AIDriveStrategyChopperCourse:updatePipeOffset(ix)
     end
     local x, _, z = localToWorld(self.storage.fruitCheckHelperWpNode.node, self.pipeOffsetX, 0, 0)
     local fieldPolygon = self.course:getFieldPolygon()
-    if not CpMathUtil.isPointInPolygon(fieldPolygon, x, z) and not self.course:isOnHeadland(ix, 1) then
+    if not CpMathUtil.isPointInPolygon(fieldPolygon, x, z) and not self.course:isOnHeadland(fruitCheckWaypoint, 1) then
         self:debug('No fruit found use but no field found use the oppisote side')
         self.pipeOffsetX = -self.pipeOffsetX
-    elseif self.course:isOnHeadland(ix, 1) then
+    elseif self.course:isOnHeadland(fruitCheckWaypoint, 1) then
             self:debug('I am on a headland enganing chase mode')
             self.pipeOffsetX = 0
             self:setPipeOffsetZ(-self.measuredBackDistance - 2)
@@ -591,10 +595,11 @@ function AIDriveStrategyChopperCourse:callUnloaderWhenNeeded()
         bestUnloader, _ = self:findUnloader(self.vehicle, nil)
         self:debug('callUnloaderWhenNeeded: stopped, need unloader here and I currently don\'t have any unloaders')
         if bestUnloader then
+            self:updatePipeOffset()
             bestUnloader:getCpDriveStrategy():call(self.vehicle, nil, self:getCurrentUnloaderIdent())
         end
         return
-    else
+    elseif not self.chaseMode then
         if not self.waypointIxWhenCallUnloader then
             self:debug('callUnloaderWhenNeeded: don\'t know yet where to meet the unloader')
             return
@@ -645,6 +650,7 @@ function AIDriveStrategyChopperCourse:callUnloaderWhenNeeded()
 end
 
 function AIDriveStrategyChopperCourse:callUnloader(bestUnloader, tentativeRendezvousWaypointIx, bestEte)
+    self:updatePipeOffset()
     if bestUnloader:getCpDriveStrategy():call(self.vehicle,
             self.course:getWaypoint(tentativeRendezvousWaypointIx), self:getNextUnloader()) then
         self.unloaderToRendezvous:set(bestUnloader, 1000 * (bestEte + 30))
@@ -697,10 +703,41 @@ function AIDriveStrategyChopperCourse.isChopper(combine)
     return capacity == math.huge
 end
 
-function AIDriveStrategyChopperCourse:headingToUpDown(ix)
-    self.connectingTrack = self.course:isOnConnectingTrack(ix)
+function AIDriveStrategyChopperCourse:getConnectingTrack()
+    return self.state == self.states.ON_CONNECTING_TRACK
 end
 
-function AIDriveStrategyChopperCourse:getConnectingTrack()
-    return self.connectingTrack
+--- We calculated a waypoint to meet the unloader (either because it asked for it or we think we'll need
+--- to unload. Now make sure that this location is not around a turn or the pipe isn't in the fruit by
+--- trying to move it up or down a bit. If that's not possible, just leave it and see what happens :)
+function AIDriveStrategyChopperCourse:findBestWaypointToUnloadOnUpDownRows(ix, isPipeInFruitAllowed)
+    local dToNextTurn = self.course:getDistanceToNextTurn(ix) or math.huge
+    local lRow, ixAtRowStart = self.course:getRowLength(ix)
+    local currentIx = self.course:getCurrentWaypointIx()
+    local newWpIx = ix
+    self:debug('Looking for a waypoint to unload around %d on up/down row, pipe in fruit %s, dToNextTurn: %d m, lRow = %d m',
+            ix, tostring(pipeInFruit), dToNextTurn, lRow or 0)
+
+    -- Waypoint would be on next row 
+    if ixAtRowStart then
+        -- so we'll have some distance for unloading
+        if dToNextTurn < AIDriveStrategyCombineCourse.safeUnloadDistanceBeforeEndOfRow then
+            local safeIx = self.course:getPreviousWaypointIxWithinDistance(ix,
+                    AIDriveStrategyCombineCourse.safeUnloadDistanceBeforeEndOfRow)
+            newWpIx = math.max(ixAtRowStart + 1, safeIx or -1, ix - 4, currentIx)
+        end
+        -- Waypoint would be on next row 
+        if ixAtRowStart > currentIx then
+            self:debug('Waypoint on next row, rejecting rendezvous')
+            newWpIx = nil
+        end
+    end
+    -- no better idea, just use the original estimated, making sure we avoid turn start waypoints
+    if newWpIx and self.course:isTurnStartAtIx(newWpIx) then
+        self:debug('Calculated rendezvous waypoint is at turn start, moving it up')
+        -- make sure it is not on the turn start waypoint
+        return math.max(newWpIx - 1, currentIx)
+    else
+        return newWpIx
+    end
 end
--- a/scripts/ai/AIDriveStrategyChopperCourse.lua
+++ b/scripts/ai/AIDriveStrategyChopperCourse.lua
@@ -169,12 +169,6 @@ function AIDriveStrategyChopperCourse:onWaypointPassed(ix, course)
     AIDriveStrategyFieldWorkCourse.onWaypointPassed(self, ix, course)
 end
 
--- function AIDriveStrategyChopperCourse:start(course, startIx, jobParameters)
---     AIDriveStrategyChopperCourse.superClass().start(self, course, startIx, jobParameters)
---     -- Update the pipeOffset side when we start work  
---     self:updatePipeOffset(startIx)
--- end
-
 function AIDriveStrategyChopperCourse:checkNextUnloader()
     if not self:getUnloader(self:getCurrentUnloaderIdent()) and self:getUnloader(self:getNextUnloader()) then
         self:debug('checkNextUnloader: I lost my current unloder and I have one that is arriving switch them')
@@ -454,9 +448,9 @@ function AIDriveStrategyChopperCourse:updatePipeOffset(ix)
     self:debug('Fruitwaypoint was set to %d', fruitCheckWaypoint)
 
     if not self.course:isOnHeadland(fruitCheckWaypoint) then
-        local lRow = self.course:getRowLength(fruitCheckWaypoint)
+        local lRow, rowStartIx = self.course:getRowLength(fruitCheckWaypoint)
         if ixAtRowStart then
-            fruitCheckWaypoint = self.course:getNextWaypointIxWithinDistance(fruitCheckWaypoint,lRow / 2)
+            fruitCheckWaypoint = self.course:getNextWaypointIxWithinDistance(startIx,lRow / 2)
             self:debug('Fruitwaypoint was set to the middle of the row %d', fruitCheckWaypoint)
         else
             fruitCheckWaypoint = fruitCheckWaypoint + 20
--- a/scripts/ai/AIDriveStrategyChopperCourse.lua
+++ b/scripts/ai/AIDriveStrategyChopperCourse.lua
@@ -53,6 +53,8 @@ function AIDriveStrategyChopperCourse.new(customMt)
         currentUnloader = CpTemporaryObject(nil),
         nextUnloader = CpTemporaryObject(nil)
     }
+
+    -- Bool to set for special logic when unloader is need to be snugged up against the chopper so we don't drive in the fruit or off the field when unloading
     self.chaseMode = false
     return self
 end
@@ -62,15 +64,13 @@ function AIDriveStrategyChopperCourse:setAllStaticParameters()
     self:checkMarkers()
     AIDriveStrategyChopperCourse.superClass().setAllStaticParameters(self)
 
+    -- Check if we are a sugarcane havester for special handling
     self.isSugarCaneHarvester = self:getSugarCaneHarvester()
 
     -- We need set this as a variable and update left/right side on turns in self:updatePipeOffset()
     self:setPipeOffsetX()
+    -- This is used to increase offset z when being chased instead along side
     self:setPipeOffsetZ()
-    
-    local total, pipeInFruit = self.vehicle:getFieldWorkCourse():setPipeInFruitMap(self.pipeOffsetX, self:getWorkWidth())
-    self:debug('Pipe in fruit map updated, there are %d non-headland waypoints, of which at %d the pipe will be in the fruit',
-            total, pipeInFruit)
 
     self:debug('AIDriveStrategyChopperCourse set')
 
@@ -78,6 +78,7 @@ end
 
 function AIDriveStrategyChopperCourse:initializeImplementControllers(vehicle)
     AIDriveStrategyChopperCourse:superClass().initializeImplementControllers(self, vehicle)
+    -- Need access to chopper controller functions
     local _
     _, self.pipeController = self:addImplementController(vehicle, PipeController, Pipe, {}, nil)
     self.combine, self.chopperController = self:addImplementController(vehicle, ChopperController, Combine, {}, nil)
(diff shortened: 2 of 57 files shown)


pops64:
Does not support the one chaff augur. First guess is the tip side setting something to look int

Tensuko:
About adding translations, pls follow our Wiki Guide.
https://github.com/Courseplay/Courseplay_FS22/wiki/Translations

Tensuko:
From my 1st test, the UI seems to fit, but I already have a few things that needs to be changed/added.

Pls remove the info msg for chopper.
info msg

Use the fruit check area to see on what side the unloader needs to drive. Also on the 1st headland, the unloader have to drive behind the chopper.
fruit check

Make sure the distance is not that high, seems like your pipe offset is not correct calculated.
distance

To solve traffic conflicts, check the combine unloader and implement the same behavior.
traffic

config/jobParameters/CombineUnloaderJobParameterSetup.xml
```diff
@@ -17,10 +17,12 @@
 			<Values>
 				<Value name="UNLOAD_COMBINE">1</Value>
 				<Value name="UNLOAD_SILO_LOADER">2</Value>
+				<Value name="UNLOAD_CHOPPER">3</Value>
```
schwiti6190, line 20: Not a big fan of adding another option here, as this complicates the usage for the player.
Maybe instead of creating a new unload copper strategy, we could keep the combine one for both but 
add a context with dependcy injection depending on the found unload target(combine or chopper)?

scripts/ai/controllers/ChopperController.lua
```diff
@@ -0,0 +1,60 @@
+---@class ChopperController : ChopperController
+
```
schwiti6190, line 2: This controller doesn't follow the convention of the other implement controllers, which are based on Spec's or external mods. Giants doesn't differentiate between combine or chopper, as both have a common combine controller for the common combine specialization. Unless you've got anything special which can be put here.

scripts/ai/controllers/CombineController.lua
```diff
@@ -124,59 +123,4 @@ end
 function CombineController:isDroppingStrawSwath()
     return self.combineSpec.strawPSenabled
 end
--------------------------------------------------------------
```
schwiti6190, line 127: See the comment above about the chopper controller.
This logic applies mostly indirect to the combine Spec and should be placed here.

config/jobParameters/CombineUnloaderJobParameterSetup.xml
pvaiko: > instead of creating a new unload copper strategy, we could keep the combine one for both 

I get why you don't like that but I still think it would be cleaner to have separate strategies. Another option could be to create a strategy for idle unloaders and instantiate the combine or chopper strategy only after the unloader is called by the combine/chopper.

scripts/ai/controllers/ChopperController.lua
pops64: I only did it to give clear separation between Chopper specific functions that CP team made to fix issues and the combine specific stuff. And as you said the combine stuff is related so I made sure to inherit the combine stuff just in case there was combine stuff used

pops64:
I will have to test for headlands. Currently relying on the pipe in fruit map for functionality. The fruit area is unreliable until your in the field. I am using the max discharge distance to calculate the pipe offset. The conflict in headland turn was an edge case I was worried about why I prefer to always drive on the not fruit side . I will have to make a special handle for headland turns. I will look at changing the text. I know the UI is a beast onto itself why I don't like to go tinkering with it.

I will fix the translations used to the manual way from FS17 days

Tensuko:
You should stop compare everything with FS17 and have a closer look at FS19 where the headland turn and traffic conflicts and side of chopper are all handeled already....

config/jobParameters/CombineUnloaderJobParameterSetup.xml
pops64: I would go further and move the unloading code out as well, just my two cents tho. You would be adding a lot of if statements to combine if you combined them as there are some things you just don't want a chopper unload driver to do

pops64:
@Tensuko I didn't play in 19 so I have to point of reference. The combine/chopper code is always a beast to look at.

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -366,35 +366,33 @@ end
...
     end
-    --TODO Add a check to see if we are on the last row and if we are set the it to field side
-    self:debug('I have left the pipeoffset on the right side')
-end
\ No newline at end of file
+    self:debug('No fruit found use the same side')
+end
+
+--- Is pipe in fruit according to the current field harvest state at waypoint?
+function AIDriveStrategyChopperCourse:isFruitAtWaypoint(course, ix, offsetX, offsetZ)
```
pvaiko, line 391: How exactly is this different than AIDriveStrategyCombineCourse:isPipeInFruitAtWaypointNow()?

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -366,35 +366,33 @@ end
...
+    local x, _, z = course:getWaypointPosition(ix)
+
+    local hasFruit = PathfinderUtil.hasFruit(x + (offsetX or 0), z + (offsetZ or 0), 1, 1)
```
pvaiko, line 394: Also, this offset calculation is incorrect.

scripts/ai/AIDriveStrategyCombineCourse.lua
```diff
@@ -897,7 +874,7 @@ function AIDriveStrategyCombineCourse:findUnloader(combine, waypoint)
             local x, _, z = getWorldTranslation(self.vehicle.rootNode)
             ---@type AIDriveStrategyUnloadCombine
             local driveStrategy = vehicle:getCpDriveStrategy()
-            if driveStrategy:isServingPosition(x, z, 0) then
+            if driveStrategy:isServingPosition(x, z, self.distanceOverFieldEdgeAllowed) then
```
pvaiko, line 877: distanceOverFieldEdgeAllowed undefined in this strategy

scripts/ai/AIDriveStrategyUnloadCombine.lua
```diff
@@ -626,8 +630,8 @@ function AIDriveStrategyUnloadCombine:getTrailersTargetNode()
                 self:debugSparse('Can\'t find trailer target node')
             end
         else
-            self:debugSparse('Combine says it can\'t load trailer')
-            --TODO: maybe then send the unloader away if activated?
+            self:debugSparse('Combine says it can\'t load trailer %s driven by %s filling %s', CpUtil.getName(trailer), CpUtil.getName(self.vehicle), g_fillTypeManager:getFillTypeByIndex(self.combineToUnload:getCpDriveStrategy():getFillType()).title)
```
pvaiko, line 633: getFillType() may return nil, and that will cause a callstack

scripts/ai/AIDriveStrategyUnloadCombine.lua
```diff
@@ -1268,7 +1272,7 @@ function AIDriveStrategyUnloadCombine:startPathfinding(
             -- the same reference everywhere
             local goal = PathfinderUtil.getWaypointAsState3D(target, -xOffset or 0, zOffset or 0)
             self.pathfinder, done, path, goalNodeInvalid = PathfinderUtil.startPathfindingFromVehicleToGoal(
-                    self.vehicle, goal, allowReverse or false, fieldNum, vehiclesToIgnore, {},
+                    self.vehicle, goal, false, fieldNum, vehiclesToIgnore, {},
```
pvaiko, line 1275: Instead of just removing this forgotten variable, we should use getAllowReversePathfinding() (useful for example for trucks)

scripts/specializations/CpAIFieldWorker.lua
```diff
@@ -354,9 +354,20 @@ function CpAIFieldWorker:replaceAIFieldWorkerDriveStrategies(jobParameters, star
         local combine = AIUtil.getImplementOrVehicleWithSpecialization(self, Combine) 
         local pipe = combine and SpecializationUtil.hasSpecialization(Pipe, combine.specializations)
         if combine and pipe then -- Default harvesters with a pipe.
-            CpUtil.debugVehicle(CpDebug.DBG_FIELDWORK, self, 'Found a combine with pipe, install CP combine drive strategy for it')
-            cpDriveStrategy = AIDriveStrategyCombineCourse.new()
-            self.spec_cpAIFieldWorker.combineDriveStrategy = cpDriveStrategy
+            local capacity = 0
```
pvaiko, line 357: This check should be implemented in the chopper strategy as an isChopper() function and not here.

Tensuko:
On first headland, we need to drive behind the chopper, again, see fs19 and proximity sensor how to keep drive behind the chopper.
grafik
The turn handling like this need way to much space.
grafik
The offset is still way too high and you see why we have to drive behind the chopper.
grafik

pops64:
We can't drive behind the combines on headland. I prefer not to alter the behavior of the game just to make CP work I prefer to make CP work to fit the behavior of the game. By this to drive behind the combine you would need to alter the code for the limit on how far away the combine will allow a discharge(I view this as cheating). Very large trailers it can't reach the back when almost full. On headlands it shouldn't be an issue cause the traffic should be turned off for the chopper when it detects the chopper(Unless it is turning) and now that it has the chopper course it shouldn't be conflicting. The problem is the outside track takes longer so it falls behind. I could possibly see if we can find the rotation of the pipe and use that instead of a hard set distance that updates based on the target(This would be complicated if at all possible).

There is no way to make the turn smaller. It relies on pathfinder which generates its own turn and then there is an alignment course at the end to help get use straight again.(Run into the chopper with out it)

My plan is to add a check for smaller work widths and adjust the offset based on that. But the turn around swings wide at the end even with the alignment course.(The hand off between states causes this issue I might have a fix for it.) Tool offset on the unloader driver should adjust this if you want it in closer.

Tensuko:
We can drive behind choppers.
The turn can be handled ON the field without any turns by pathfinder.
The offset can be adjusted based on the pipe.

Without those points it will be a downgrade compared to FS19 and I will not merge this into FS22.

pops64:
We can't without cheating. Fill a trailer about around 80 to 90% with the longest chaff trailler in the shop and try driving behind a chopper. Make sure you do it on a clean game with no CP(The code that enables this cheat doesn't reset on CP being stopped). We had this issue in 17 but we also didn't adjust for the target node moving in the trailer. I don't mind adjusting the game when CP interacts faster than what Giants intended but when a human player physical can't do the same thing as CP, I won't code that into CP.

The turn can be handled be AP but with reduce results(Think of AD drivers chopper unloading currently).

Pipeoffset can not be used. It returns zero or mixed results. Maxdischarge distance has to be used.

And currently nothing will be a downgrade. FS22 as been out for 3 years now and a core feature that should have never been abandoned is still MIA. So either merge or not but until you link to the code you think I should change and provide code suggestions. Deal with CP like we did in 17, if CP can't handle it do it your self.

pops64:
Due to differing interests this will no longer be added back to CP unless it is done by the CP owner. I am no longer interested in this support so this will be closed. I will leave my fork this active for those that want to merge my code base into yours. It was mostly feature complete needed some minor tweaks and some logic for better handling of avoiding fruit when departing on a full load is missing but it will unload the chopper just fine. It can't handle sugarcane unfortunately.

pops64:
I am going to reopen this I have put to much work into this for me to be the one to close it out.

pops64:
Only active on the first row and the first row of a land

scripts/ai/AIDriveStrategyCombineCourse.lua
```diff
@@ -1747,35 +1674,11 @@ end
 --- Deregister a combine unload AI driver from notifications
 ---@param driver AIDriveStrategyUnloadCombine
 function AIDriveStrategyCombineCourse:deregisterUnloader(driver, noEventSend)
+    self:debug('Unloader has be unregistered')
```
pvaiko, line 1677: typo

scripts/ai/AIDriveStrategyUnloadCombine.lua
```diff
@@ -1272,11 +1272,11 @@ function AIDriveStrategyUnloadCombine:startPathfinding(
             -- the same reference everywhere
             local goal = PathfinderUtil.getWaypointAsState3D(target, -xOffset or 0, zOffset or 0)
             self.pathfinder, done, path, goalNodeInvalid = PathfinderUtil.startPathfindingFromVehicleToGoal(
-                    self.vehicle, goal, allowReverse or false, fieldNum, vehiclesToIgnore, {},
+                    self.vehicle, goal, self:getAllowReversePathfinding() or false, fieldNum, vehiclesToIgnore, {},
```
pvaiko, line 1275: or false unnecessary

scripts/ai/AIDriveStrategyUnloadCombine.lua
```diff
@@ -1272,11 +1272,11 @@ function AIDriveStrategyUnloadCombine:startPathfinding(
...
                     maxFruitPercent, self.offFieldPenalty, areaToAvoid)
         else
             self.pathfinder, done, path, goalNodeInvalid = PathfinderUtil.startPathfindingFromVehicleToNode(
-                    self.vehicle, target, xOffset or 0, zOffset or 0, false,
+                    self.vehicle, target, xOffset or 0, zOffset or 0, self:getAllowReversePathfinding() or false,
```
pvaiko, line 1279: or false unnecessary

scripts/ai/AIDriveStrategyUnloadCombine.lua
```diff
@@ -2219,8 +2219,13 @@ function AIDriveStrategyUnloadCombine:onPathfindingDoneBeforeUnloadingOnField(pa
                 -self.fieldUnloadData.xOffset, dz, zOffset, 3, false))
         self:startCourse(course, 1)
     else
-        self:info("Could not find a path to the field unload position!")
-        self.vehicle:stopCurrentAIJob(AIMessageCpErrorNoPathFound.new())
+        -- This lets pathfinder try again with less strict pathfinder instead of just outright going home with its ball
```
pvaiko, line 2222: Please do not change existing logic, revert this change. @schwiti6190 We'll need to review this whole pathfinder retry logic anyway as it does not make a lot of sense as it is now. It'll relax the constraints but at the next try, the situation will most likely be different, a failed pathfinder run takes a long time and in that time the harvester most likely moved .

scripts/ai/controllers/PipeController.lua
```diff
@@ -284,6 +287,14 @@ function PipeController:measurePipeProperties()
             self.implement, self.implement.rootNode)
         self.pipeOffsetZ = pipeOffsetZ + dist
     end
+    if ImplementUtil.isChopper(self.implement) then
```
pvaiko, line 290: Why do we need this? The existing pipe measurement does not work?

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
+--[[
+This file is part of Courseplay (https://github.com/Courseplay/courseplay)
+Copyright (C) 2021 Peter Vaiko
```
pvaiko, line 3: date, author

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
+--[[
+This file is part of Courseplay (https://github.com/Courseplay/courseplay)
```
pvaiko, line 2: fix link

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+
+    self:debug('AIDriveStrategyChopperCourse set')
+
+end
+
+function AIDriveStrategyChopperCourse:initializeImplementControllers(vehicle)
+    AIDriveStrategyChopperCourse:superClass().initializeImplementControllers(self, vehicle)
+    -- Need access to chopper controller functions
+    local _
+    _, self.pipeController = self:addImplementController(vehicle, PipeController, Pipe, {}, nil)
```
pvaiko, line 83: The parent already added a PipeController.

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    self.combine, self.chopperController = self:addImplementController(vehicle, ChopperController, Combine, {}, nil)
```
pvaiko, line 84: assigning self.combine here is suspect. Already assigned in parent.

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+        nextUnloader = CpTemporaryObject(nil)
+    }
+
+    -- Bool to set for special logic when unloader is need to be snugged up against the chopper so we don't drive in the fruit or off the field when unloading
+    self.chaseMode = false
+    return self
+end
+
+function AIDriveStrategyChopperCourse:setAllStaticParameters()
+    --Old Code to add markers to Choppers which don't have AI markers
```
pvaiko, line 63: Old code? What does this mean?

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    self.combine, self.chopperController = self:addImplementController(vehicle, ChopperController, Combine, {}, nil)
+end
+
+--------------------------------------------------------------------------------------------------------------------------------------------------------------
+-- Main Loop
+---------------------------------------------------------------------------------------------------------------------------------------------------------------
+
+function AIDriveStrategyChopperCourse:update(dt)
+    AIDriveStrategyCombineCourse.update(self, dt)
+    -- Old Code need to make Choppers work
```
pvaiko, line 93: Old code? What does this mean?

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    if self.temporaryHold:get() then
+        self:setMaxSpeed(0)
+    end
+    if self.state == self.states.WORKING then
+        -- Harvesting
+        self:checkRendezvous()
+        self:checkBlockingUnloader()
+
+        if self:isChopperWaitingForUnloader() then
+            self:debug('No trailer calling for an unloader')
```
pvaiko, line 109: Don't understand debug message

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+            self:stopForUnload(self.states.WAITING_FOR_UNLOAD_ON_FIELD, true)
+        end
+
+    elseif self.state == self.states.TURNING then
+        self:checkBlockingUnloader()
+    elseif self.state == self.states.WAITING_FOR_LOWER then
+        if self:isChopperWaitingForUnloader() then
+            self:debug('No trailer calling for an unloader')
```
pvaiko, line 117: Don't understand debug message

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+            self:stopForUnload(self.states.WAITING_FOR_UNLOAD_ON_FIELD, true)
+        end
+    elseif self.state == self.states.UNLOADING_ON_FIELD then
+        -- Unloading
+        self:driveUnloadOnField()
```
pvaiko, line 122: driveUnloadOnField does not make sense for a chopper as it is always unloading. This state and self.unloadState should be eliminated.

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    self:updatePipeOffset(ix)
+    AIDriveStrategyChopperCourse.superClass().resumeFieldworkAfterTurn(self, ix)
+end
+
+-----------------------------------------------------------------------------------------------------------------------
+--- Event listeners
+-----------------------------------------------------------------------------------------------------------------------
+function AIDriveStrategyChopperCourse:onWaypointPassed(ix, course)
+    if self.state == self.states.UNLOADING_ON_FIELD and
+            (self.unloadState == self.states.DRIVING_TO_SELF_UNLOAD or
```
pvaiko, line 140: driving to self unload? Remove obsolete states.

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+        if lastPassedWaypointIx > self.unloaderRendezvousWaypointIx then
+            -- past the rendezvous waypoint
+            self:debug('Unloader missed the rendezvous at %d', self.unloaderRendezvousWaypointIx)
+            local unloaderWhoDidNotShowUp = self.unloaderToRendezvous:get()
+            -- need to call this before onMissedRendezvous as the unloader will call back to set up a new rendezvous
+            -- and we don't want to cancel that right away
+            self:cancelRendezvous()
+            unloaderWhoDidNotShowUp:getCpDriveStrategy():onMissedRendezvous(self.vehicle)
+        end
+        if self:getNextUnloader() and self:getNextUnloader():readyToRecive() then
```
pvaiko, line 172: typo (Recive)

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+        self:estimateDistanceUntilFull(ix)
+        self:callUnloaderWhenNeeded()
+        -- Alteration from parent function we need to check every once in a while to see if the next unloader incoming is ready to replace out current unloader
+        self:checkNextUnloader()
+    end
+
+    AIDriveStrategyFieldWorkCourse.onWaypointPassed(self, ix, course)
+end
+
+function AIDriveStrategyChopperCourse:checkRendezvous()
```
pvaiko, line 160: Needs explanation here, what a rendezvous means here, as obviously, with a chopper, an unloader always must be present, not like with a combine which can arrange to meet an unloader in the future

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+        end
+        if self:getNextUnloader() and self:getNextUnloader():readyToRecive() then
+            -- We only cancel the rendezvous if we try to switch the unloaders here there will be mutliple calls to it causing logic breaks
+            self:debug('Discharging to %s, cancelling unloader rendezvous %s is ready to come along side', CpUtil.getName(self:getCurrentUnloader().vehicle), CpUtil.getName(self:getNextUnloader().vehicle))
+            self:cancelRendezvous()
+        end
+    end
+end
+
+function AIDriveStrategyChopperCourse:driveUnloadOnField()
```
pvaiko, line 180: As above, this makes no sense for a chopper

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+        end
+        self:setMaxSpeed(0)
+    end
+end
+
+-----------------------------------------------------------------------------------------------------------------------
+--- Pipe handling
+-----------------------------------------------------------------------------------------------------------------------
+
+-- This part of an ugly workaround to make the chopper pickups work
```
pvaiko, line 207: This is from FS19 and must be verified if still valid and needed

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+end
+
+function AIDriveStrategyChopperCourse:setPipeOffsetZ(offset)
+    -- If offset is supplied use that(Chase Mode) plus what is returned from pipecontroller
+    self.pipeOffsetZ = (offset or 0) + self.pipeController:getPipeOffsetZ()
+end
+
+-- Return our current pipe offset plus any user supplied offset
+function AIDriveStrategyChopperCourse:getPipeOffset(additionalOffsetX, additionalOffsetZ)
+    self:debugSparse('Chopper PipeOffsetX is %.2f', self.pipeOffsetX)
```
pvaiko, line 249: what's the point of this debug it does not show the resulting offset

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    self:debug('Pipe Offset X was set as: %.2f', self.pipeOffsetX)
+end
+
+function AIDriveStrategyChopperCourse:setPipeOffsetZ(offset)
+    -- If offset is supplied use that(Chase Mode) plus what is returned from pipecontroller
+    self.pipeOffsetZ = (offset or 0) + self.pipeController:getPipeOffsetZ()
+end
+
+-- Return our current pipe offset plus any user supplied offset
+function AIDriveStrategyChopperCourse:getPipeOffset(additionalOffsetX, additionalOffsetZ)
```
pvaiko, line 248: Is this guaranteed to never be called with nil?

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    self:debugSparse('Chopper PipeOffsetX is %.2f', self.pipeOffsetX)
+    return self.pipeOffsetX + additionalOffsetX, self.pipeOffsetZ + additionalOffsetZ
+end
+
+function AIDriveStrategyChopperCourse:updatePipeOffset(ix)
```
pvaiko, line 253: Pipe offset for a chopper does not make sense. I understand you are trying to use that as some kind of target for the unloader, but 1) I prefer the unloader deciding where to drive and not the chopper telling it 2) but at least, it must not be called pipeOffset as it is misleading and confusing, rather unloaderTargetPosition or something like that.

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    -- If we are on the first headland engage chase mode and the pipeoffset to zero disabled for sugarcane
+    if self.course:isOnHeadland(fruitCheckWaypoint, 1 ) and not self.isSugarCaneHarvester then
+        self:debug('I am on a headland enganing chase mode')
+        self.pipeOffsetX = 0
+        self:setPipeOffsetZ(-self.measuredBackDistance - 2) 
+        -- We need this so the unloader knows what turn manuvers to use
+        self.chaseMode = true
+    else
+        -- Check to see if the waypoint that we are using would put the unloader in the fruit using out current pipe offset
+        -- And if would use the oppisite vaule of the pipe offset as the should be cleared because the should be our last row worked
```
pvaiko, line 292: typos

scripts/ai/AIDriveStrategyUnloadChopper.lua
```diff
@@ -0,0 +1,867 @@
+--[[
+This file is part of Courseplay (https://github.com/Courseplay/Courseplay_FS22)
+Copyright (C) 2022 Peter Vaiko
```
pvaiko, line 3: date, author

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    end
+    
+    -- We need a waypoint to check use a supplied one(When we are getting the next unloader) or if nothing is supplied just use our current one
+    local fruitCheckWaypoint = ix or self.course:getCurrentWaypointIx()
+    self:debug('updatePipeOffset: Fruitwaypoint was set to %d', fruitCheckWaypoint)
+
+    if not self.course:isOnHeadland(fruitCheckWaypoint) then
+        local lRow, rowStartIx = self.course:getRowLength(fruitCheckWaypoint)
+        if rowStartIx then
+            fruitCheckWaypoint = self.course:getNextWaypointIxWithinDistance(rowStartIx, lRow / 2)
```
pvaiko, line 275: why middle of the row? that can be in front of us or behind us.

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+            fruitCheckWaypoint = fruitCheckWaypoint + 10
+            self:debug('Fruitwaypoint was set 10 ahead to couldn\'t determine row length %d', fruitCheckWaypoint)
+        end
+    end
+
+    -- If we are on the first headland engage chase mode and the pipeoffset to zero disabled for sugarcane
+    if self.course:isOnHeadland(fruitCheckWaypoint, 1 ) and not self.isSugarCaneHarvester then
+        self:debug('I am on a headland enganing chase mode')
+        self.pipeOffsetX = 0
+        self:setPipeOffsetZ(-self.measuredBackDistance - 2)
```
pvaiko, line 287: why -2? what exactly -2 mean? What is going to be 2 meters from the back?

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+                self.chaseMode = true
+                self.landRow = true
+                return
+            end
+        end
+        
+        -- Once final check to make sure the pipeoffset we are going to supply to the unloader is on the field.
+        local x, _, z = localToWorld(self.storage.fruitCheckHelperWpNode.node, self.pipeOffsetX, 0, 0)
+        local fieldPolygon = self.course:getFieldPolygon()
+        if not CpMathUtil.isPointInPolygon(fieldPolygon, x, z) then
```
pvaiko, line 315: Should use one of the CpFieldUtil functions (like isNodeOnField), isPointInPolygon is an expensive call.

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+            -- The point we are using isn't on the field check the otherside of the chopper to see if there is fruit and if there is no fruit use have the unloader follow us on that side
+            -- Otherwise just have the unloader follow us
+            if hasFruit and not self.isSugarCaneHarvester then
```
pvaiko, line 318: Also, what case ist this even? Not in the field but there is fruit? On an adjacent field or what?

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    return self.unloaders.nextUnloader:get()
+end
+
+-- Another saftey check to make sure we always have a current unloader
+-- This is ran every waypointpassed() call. It asks if our next unloader is behind and aligned and within 25 meters of lastwaypoint of its driveToCombine Course
+function AIDriveStrategyChopperCourse:checkNextUnloader()
+    if not self:getCurrentUnloader() and self:getNextUnloader() then
+        self:debug('checkNextUnloader: I lost my current unloder and I have one that is arriving switch them')
+        self:updateNextUnloader()
+    elseif self:getNextUnloader() and self:getNextUnloader():readyToRecive() then
```
pvaiko, line 427: Recive

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    local workWidth = self:getWorkWidth()
+    local x, _, z = localToWorld(self.vehicle:getAIDirectionNode(), workWidth, 0, 0)
+    self.fieldOnLeft = CpFieldUtil.isOnField(x, z)
+    x, _, z = localToWorld(self.vehicle:getAIDirectionNode(), -workWidth, 0, 0)
+    self.fieldOnRight = CpFieldUtil.isOnField(x, z)
+    self:debug('Fruit left: %.2f right %.2f, field on left %s, right %s',
+            self.fruitLeft, self.fruitRight, tostring(self.fieldOnLeft), tostring(self.fieldOnRight))
+end
+
+-- Function to allow us to revice the nearest trailer objects
```
pvaiko, line 715: typo

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+function AIDriveStrategyChopperCourse:nearestChopperTrailer()
```
pvaiko, line 716: nearest is misleading. This is the trailer we are currently discharging into

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    local trailer = self.pipeController:getClosestObject()
+    local targetObject = self.pipeController:getDischargeObject()
+    return trailer, targetObject
+end
+
+-- Return true for no headlands so we can reduce off field pently for our unloader driver
```
pvaiko, line 722: typo

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    self:debug('Will be full at waypoint %d, fill level %d at waypoint %d (current waypoint %d), %.1f m and %.1f l until call (currently %.1f l), wp distance %.1f',
+            self.waypointIxWhenFull or -1, self.callUnloaderAtFillLevelPercentage, self.waypointIxWhenCallUnloader or -1,
+            self.course:getCurrentWaypointIx(), dUntilCallUnloader, litersUntilCallUnloader, fillLevel, wpDistance)
+end
+
+---------------------------------------------------------------------------------------------------------------------------------------------------------
+-- Utilites
+-----------------------------------------------------------------------------------------------------------------------------------------------------
+
+-- Check for sugarcane havesert certian logic needs to be disabled when we are unloding them
```
pvaiko, line 677: typo, typo, typo

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    self:debugSparse('isReadyToUnload(): defaulting to ready to unload')
+    return true
+end
+
+function AIDriveStrategyChopperCourse:getConnectingTrack()
+    return self.state == self.states.ON_CONNECTING_TRACK
+end
+
+
+-- Copied from parent altered to remove pipe in fruit map logic. We don't have a pipeInFruit map for choppers
```
pvaiko, line 599: why not?

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    elseif self:getNextUnloader() and self:getNextUnloader():readyToRecive() then
+        self:debug('checkNextUnloader: Discharging to %s, and %s is ready to come along side', CpUtil.getName(self:getCurrentUnloader().vehicle), CpUtil.getName(self:getNextUnloader().vehicle))
+        --TODO This is the user method to request drive now. It resets after 3 secs which may cause and already full unloader to return if there is a reason for it go to idle state from drive to unload
+        self:getCurrentUnloader():requestDriveUnloadNow()
+        self:updateNextUnloader()
+    end
+end
+
+
+-- Coppied from parent. Altered to handle a current and next unloader
```
pvaiko, line 436: typo

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+function AIDriveStrategyChopperCourse:callUnloaderWhenNeeded()
```
pvaiko, line 437: This has way too many lines common with the parent, if that really is needed, it should be split so the business logic of finding a rendezvous point is not duplicated.

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+                -- do not call too early (like minutes before we get there), only when it needs at least as
+                -- much time to get there as the combine (-5 seconds)
+                self:callUnloader(bestUnloader, tentativeRendezvousWaypointIx, bestEte)
+            end
+        end
+    end
+end
+
+-- Copied from parent altered to check for active chopper unloaders
+function AIDriveStrategyChopperCourse:findUnloader(combine, waypoint)
```
pvaiko, line 516: same here, way too many common lines, need a more intelligent solution to not repeat business logic code.

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    if bestUnloader then
+        self:debug('findUnloader: best unloader is %s (score %.1f, ETE %.1f)',
+                CpUtil.getName(bestUnloader), bestScore, bestEte)
+        return bestUnloader, bestEte
+    else
+        self:debugSparse('findUnloader: no idle unloader found')
+    end
+end
+
+-- Copied from parent removed pocket logic and other things that were causes logic issues
```
pvaiko, line 560: grammar

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    if self.course:isCloseToNextTurn(10) then
+        self:debugSparse('isReadyToUnload(): too close to turn')
+        return false
+    end
+    -- safe default, better than block unloading
+    self:debugSparse('isReadyToUnload(): defaulting to ready to unload')
+    return true
+end
+
+function AIDriveStrategyChopperCourse:getConnectingTrack()
```
pvaiko, line 594: isOnConnectingTrack() would better describe what this means.

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    return self.state == self.states.ON_CONNECTING_TRACK
+end
+
+
+-- Copied from parent altered to remove pipe in fruit map logic. We don't have a pipeInFruit map for choppers
+--- We calculated a waypoint to meet the unloader (either because it asked for it or we think we'll need
```
pvaiko, line 600: Comment must be updated

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+--- to unload. Now make sure that this location is not around a turn or the pipe isn't in the fruit by
+--- trying to move it up or down a bit. If that's not possible, just leave it and see what happens :)
+function AIDriveStrategyChopperCourse:findBestWaypointToUnloadOnUpDownRows(ix, isPipeInFruitAllowed)
```
pvaiko, line 603: function name does not seem to describe what this is used for here (finding a waypoint to switch unloaders)

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    local dToNextTurn = self.course:getDistanceToNextTurn(ix) or math.huge
```
pvaiko, line 604: Also, does this even apply for the switch unloaders use case?

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+        -- make sure it is not on the turn start waypoint
+        return math.max(newWpIx - 1, currentIx)
+    else
+        return newWpIx
+    end
+end
+
+
+-- Copied from parent altered to use the trailer fill level instead of the chopper
+function AIDriveStrategyChopperCourse:estimateDistanceUntilFull(ix)
```
pvaiko, line 632: Here again, way too much common code with the parent, need a smarter way to do this (like passing in a lambda to get the fill level and capacity) to avoid repeating the same business logic.

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+            self.waypointIxWhenFull or -1, self.callUnloaderAtFillLevelPercentage, self.waypointIxWhenCallUnloader or -1,
+            self.course:getCurrentWaypointIx(), dUntilCallUnloader, litersUntilCallUnloader, fillLevel, wpDistance)
+end
+
+---------------------------------------------------------------------------------------------------------------------------------------------------------
+-- Utilites
+-----------------------------------------------------------------------------------------------------------------------------------------------------
+
+-- Check for sugarcane havesert certian logic needs to be disabled when we are unloding them
+function AIDriveStrategyChopperCourse:getSugarCaneHarvester()
```
pvaiko, line 678: isSugarCaneHarvester() is what this should be called

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+    for i, fillUnit in ipairs(self.vehicle:getFillUnits()) do
+        if self.vehicle:getFillUnitSupportsFillType(i, FillType.SUGARCANE) then
+            self:debug('This is a Sugarcane havester')
```
pvaiko, line 681: surprise, surprise, another typo...

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+            return true
+        end
+    end
+    return false
+end
+
+
+-- Not what this does
```
pvaiko, line 689: Unintelligible comment

scripts/ai/AIDriveStrategyChopperCourse.lua
```diff
@@ -0,0 +1,782 @@
...
+function AIDriveStrategyChopperCourse:checkFruit()
```
pvaiko, line 690: How is this different from the current function in the parent?

scripts/ai/AIUtil.lua
```diff
@@ -324,6 +324,8 @@ end
 function AIUtil.getFirstAttachedImplement(vehicle,suppressLog)
 	-- by default, it is the vehicle's front
 	local maxDistance = vehicle.size.length / 2 + vehicle.size.lengthOffset
+	local _, _, d = localToLocal(vehicle.rootNode, AIUtil.getDirectionNode(vehicle), 0, 0, 0)
+	maxDistance = maxDistance + rootNodeOffset
```
pvaiko, line 330: rootNodeOffset undefined, copy paste error?

pops64:
So I am very close to finalizing this. I just need some testers to put it through its paces. There is one end of work bug. I think I have squashed it. There is also a combineToUnload being called when nil. It isn't game breaking but generates that ugly block of log text once. I think it as to do with checkRendevous but the log isn't providing any hints to where it is at. 

Let me know if there are oddities or rough spots still left. I know the turns are still large but it is using pathfinder so if there are any obstacles it will avoid them as best as pathfinder can do

coatsy35:
> So I am very close to finalizing this. I just need some testers to put it through its paces. There is one end of work bug. I think I have squashed it. There is also a combineToUnload being called when nil. It isn't game breaking but generates that ugly block of log text once. I think it as to do with checkRendevous but the log isn't providing any hints to where it is at.
> 
> Let me know if there are oddities or rough spots still left. I know the turns are still large but it is using pathfinder so if there are any obstacles it will avoid them as best as pathfinder can do

I can help test this if you want.

regs1980:
> > I can help test this if you want.

Test and write it here your experience
https://github.com/Courseplay/Courseplay_FS22/issues/2701

Tensuko:
https://github.com/Courseplay/Courseplay_FS22/issues/2701#issuecomment-1705988248

coatsy35:
That's a shame. CP is in need of chopper mode. Just about to go into the phase of my farm, where I'm going to have a Biogas plant and large cow sheds, so would be chopping a lot. Guess I'll be using AD. @pops64 is it impossible for your vision to adhere to the CP coding standards. Lots of people would benefit from your code and knowledge on chopping I'm sure. The chopper presumably would have been in a separate module, so it was not confused with harvesting etc...

Tensuko:
We are not in need of a chopper mode.
Nothing is that urgent to have an uncomplete and "not so well" working mode just to rush something.

Tensuko:
Addition: This is not a "there will never be a chopper mode".
But not right now, when there is time for it, Our priorities right now are on other projects as you might know.

coatsy35:
> Addition: This is not a "there will never be a chopper mode". But not right now, when there is time for it, Our priorities right now are on other projects as you might know.

I understand. I'm mostly looking forward to the day when we have new course generator. It looks ace from the screenshots I've seen 😍

pops64:
I am done being nice Tensuko. The current CP owner(Tensuko) would rather disfranchise the community then include code that works well and was still under development just because it currently doesn't work they he wants it. It was a sad day when he put lead of CP back in my days of developing CP in FS17 the CP team not any one individual made decisions on whether to include code and we usually did agree. For those that have been waiting 3 years for a core feature that has been apart of CP since at least FS15 when silage was introduced I will leave my fork open. Please post any game breaking bugs there. Any further development has been halted but I will fix crashes and keep it updated with the main branch. https://github.com/pops64/Courseplay_FS22/tree/ChopperSupport

PR #2769, opened 2023-09-14
Refactored debugs and added console commands

schwiti6190:
New console commands for printing settings:
- cpSettingsPrintVehicle
- cpSettingsPrintGenerator
- cpSettingsPrintJob
- cpSettingsPrintGlobal

### DIFF
--- a/scripts/CpUtil.lua
+++ b/scripts/CpUtil.lua
@@ -109,30 +109,6 @@ function CpUtil.getVariable(variableName)
 	return f and f() or nil
 end
 
-function CpUtil.internalPrint(printFunc, infoString, ...)
-	CpUtil.try(
-		function (...)
-			printFunc(string.format('%s %s', infoString, string.format(...)))
-		end,
-		...)
-end
-
--- convenience debug function that expects string.format() arguments,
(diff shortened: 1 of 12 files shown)


scripts/CpUtil.lua
```diff
@@ -149,6 +125,35 @@ function CpUtil.getName(object)
...
+	str = string.format("%s: %s", CpUtil.getName(implement), str)
+	CpUtil.internalPrintVehicle(implement.rootVehicle, tag, str, ...)
+end
+
+-- convenience debug function that expects string.format() arguments,
+-- CpUtil.debugVehicle( CpDebug.DBG_TURN, "fill level is %.1f, mode = %d", fillLevel, mode )
+---@param channel number
+function CpUtil.debugFormat(channel, ...)
+	if CpDebug and CpDebug:isChannelActive(channel) then
+		CpUtil.internalPrint("dbg"..channel, ...)
```
pvaiko, line 153: Since we are already at it, don't we want to get rid of this channel as a number, which isn't explicitly defined anywhere, I have no idea which number is which? Or replace it with short, meaningful string?

PR #2785, opened 2023-09-20
Feature/improve harvester turning

IhanaMies:
Due to distance between the front implement and vehicle center, some vehicles don't have enough time to align themselves properly for a new row and the vehicle does few aligning adjustments post-turn missing some crops.

Another small improvement included is turning off the cutter when turning.

(I'm marking this as a draft for futher discussion and approval)

The cases shown were done using the Ventor 4150 potato harvester.

Fig 1. Currently generated turn.
Farming Simulator 22 Screenshot 2023 09 20 - 22 01 00 76

Fig 2. After the turn, front implement is very close to field border.
Farming Simulator 22 Screenshot 2023 09 20 - 22 01 20 01

Fig 3. Harvester missing some crops still aligning itself post-turn.
Farming Simulator 22 Screenshot 2023 09 20 - 22 01 28 30

Fig 4. and 5. This is more noticeable when the Ventor 4150 does a K-turn when reversing is enabled.
(This case was done with 3 convoy setting and skipping one lane to avoid collision)
Farming Simulator 22 Screenshot 2023 09 20 - 23 13 51 22
Farming Simulator 22 Screenshot 2023 09 20 - 23 14 13 30

By taking into consideration the frontMarkerDistance, we can easily give the harvester more space to adjust properly.
Ventor 4150 has a very long spacing to the front implement which gives a lot of space. Maybe cap the max value?

Fig 6. Generated turn after setting goalOffSet to negative frontMarkerDistance.
Farming Simulator 22 Screenshot 2023 09 20 - 21 59 55 40

Fig 7. Front implement had a nice spacing to row start and harvester had time to do final adjustments.
Farming Simulator 22 Screenshot 2023 09 20 - 21 54 07 03

Fig 8. No missed crops.
Farming Simulator 22 Screenshot 2023 09 20 - 22 03 44 03

Fig 9. New reverse enabled turn on 3 convoy, 1 lane skip setting
(Reversing in this case might be redundant and a bit silly looking)
Farming Simulator 22 Screenshot 2023 09 20 - 23 22 21 29

Fig 10. Properly aligned
Farming Simulator 22 Screenshot 2023 09 20 - 23 22 46 74

Fig 11. New K-turn with frontMarkerDistance considered. Single tool. No skips.
Length between the implement and vehicle makes the reversing distance quite large. Might need to check this further.
Farming Simulator 22 Screenshot 2023 09 20 - 23 47 12 79

### DIFF
--- a/scripts/ai/AIDriveStrategyCombineCourse.lua
+++ b/scripts/ai/AIDriveStrategyCombineCourse.lua
@@ -537,7 +537,6 @@ end
 
 --- Some of our turns need a short look ahead distance, make sure we restore the normal after the turn
 function AIDriveStrategyCombineCourse:resumeFieldworkAfterTurn(ix)
-    self.state = self.states.WORKING
     self.ppc:setNormalLookaheadDistance()
     AIDriveStrategyCombineCourse.superClass().resumeFieldworkAfterTurn(self, ix)
 end
--- a/scripts/ai/AIDriveStrategyCombineCourse.lua
+++ b/scripts/ai/AIDriveStrategyCombineCourse.lua
@@ -537,6 +537,7 @@ end
 
 --- Some of our turns need a short look ahead distance, make sure we restore the normal after the turn
 function AIDriveStrategyCombineCourse:resumeFieldworkAfterTurn(ix)
+    self.state = self.states.WORKING
     self.ppc:setNormalLookaheadDistance()
     AIDriveStrategyCombineCourse.superClass().resumeFieldworkAfterTurn(self, ix)
 end
@@ -1307,7 +1308,7 @@ end
 
 --- Can the cutter be turned off ?
 function AIDriveStrategyCombineCourse:getCanCutterBeTurnedOff()
-    return self:isWaitingForUnload() or
+    return self:isWaitingForUnload() or self.state == self.states.TURNING or
             (self.state == self.states.UNLOADING_ON_FIELD and self:isUnloadStateOneOf(self.selfUnloadStates) and
             -- we want that cutter to be turned on when returning to fieldwork after self unload
                     self.unloadState ~= self.states.RETURNING_FROM_SELF_UNLOAD)
--- a/scripts/ai/turns/AITurn.lua
+++ b/scripts/ai/turns/AITurn.lua
@@ -591,6 +591,8 @@ function CourseTurn:endTurn(dt)
         if not self.implementsLowered then
             -- have not started lowering implements yet
             self:debug('Turn ending, lowering implements')
+            -- driveStrategy.state can be set to WORKING which is also checked if harvester header can be turned on
+            self.driveStrategy.state = self.driveStrategy.states.WORKING
             self.driveStrategy:lowerImplements()
             self.implementsLowered = true
             if self.ppc:isReversing() then
--- a/scripts/ai/turns/TurnContext.lua
+++ b/scripts/ai/turns/TurnContext.lua
@@ -458,7 +458,10 @@ function TurnContext:getTurnEndNodeAndOffsets(steeringLength)
         -- implement in front of vehicle. Turn should end with the implement at the work start position, this is where
         -- the vehicle's root node is on the vehicleAtTurnEndNode
         turnEndNode = self.vehicleAtTurnEndNode
-        goalOffset = 0
+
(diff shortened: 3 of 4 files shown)


scripts/ai/AIDriveStrategyCombineCourse.lua
```diff
@@ -1307,7 +1307,7 @@ end
 
 --- Can the cutter be turned off ?
 function AIDriveStrategyCombineCourse:getCanCutterBeTurnedOff()
-    return self:isWaitingForUnload() or
+    return self:isWaitingForUnload() or self.state == self.states.TURNING or
```
pvaiko, line 1310: This change would most likely turn off the cutters for normal combines making a headland turn with/without a pocket.

scripts/ai/turns/AITurn.lua
```diff
@@ -591,6 +591,8 @@ function CourseTurn:endTurn(dt)
         if not self.implementsLowered then
             -- have not started lowering implements yet
             self:debug('Turn ending, lowering implements')
+            -- driveStrategy.state can be set to WORKING which is also checked if harvester header can be turned on
+            self.driveStrategy.state = self.driveStrategy.states.WORKING
```
pvaiko, line 595: It is a major no-no to set a private variable from outside the class. Strategy state must not be changed from a different class.

pvaiko: (especially not such an apparently hacky way to turn the cutter on through side effects)

scripts/ai/turns/TurnContext.lua
```diff
@@ -458,7 +458,10 @@ function TurnContext:getTurnEndNodeAndOffsets(steeringLength)
         -- implement in front of vehicle. Turn should end with the implement at the work start position, this is where
         -- the vehicle's root node is on the vehicleAtTurnEndNode
         turnEndNode = self.vehicleAtTurnEndNode
-        goalOffset = 0
+
+        -- set the turning endpoint a bit futher from the field starting point
+        -- for better aligment for front implements (e.g. potatoharvester)
+        goalOffset = -self.frontMarkerDistance
```
pvaiko, line 464: Why the frontMarkerDistance? The turnEndNode is already frontMarkerDistance meters back from the work start. I agree some vehicles may need to finish the turn a bit further back but:
1. even with the potato harvester, front marker distance is 8-10 m or so, so this makes every turn 8-10 m bigger, causing problems when there are obstacles around the field
2. we have a similar problem with towed implements, and the issue is the same as the TODO further down in the code suggests (_TODO: this isn't exact science here, as the distance we need to straighten out the implement is rather a function of the radius, the starting angle and probably the tow bar length._) The extra distance we need is certainly not simply the front marker distance.

So here we'd need something that does not automatically increase the size of the turn for every vehicle and for those it increases, it does only as much as needed. 

I was too lazy to figure out the geometry for this, probably that's why we left it as is. I think, besides the front marker distance, a calculation should take into account the working width, the turning radius and probably even the depth of the work area.

pvaiko (changes requested):
I'd suggest to address turning off the cutter in a different PR, and focus this one on the turn only.

scripts/ai/turns/TurnContext.lua
IhanaMies: > Why the frontMarkerDistance?
It seemed like a good place to start. And yes, it makes every turn bigger and in case of potato harvester, a lot bigger. I made test capping the value to -4.0 meters which makes it less bigger, but still aligning good enough not to miss any crops.

For a very big machine like the potato harvester it is a fact it needs a lot of space to turn and without reversing enable or making headlands, it does extend quite far which I think is up to the user to set up the field/course properly.

This does make it much better for fields which does have enough space around them.

scripts/ai/AIDriveStrategyCombineCourse.lua
IhanaMies: True. Good catch. I will test this throughoutly. First intuition would be to make row changing turn it's own state, but I need to check how much that would affect other code.

scripts/ai/turns/TurnContext.lua
Tensuko: With turn on field on + lower ealry it doesn't miss a single fruit already.

pvaiko:
Closing due to inactivity.

PR #2864, opened 2023-10-18
Update translation_it.xml

Roby1164:


Tensuko:
You may want to check if there is enough space on the hud for your Shield height offset translation :)

Roby1164:
At this moment I can't check if it enters the Hud, in game.
The problem is that the translation of "offset" is position compensation, I don't know how else to describe it. Or you could abbreviate it to "Compensation." Otherwise we leave the English wording "Offset" hoping that everyone will understand it

Tensuko:
And something like shield hight adjust?
What would that be?

Roby1164:
Let's leave "Offset", since it is a voice understandable by many. At the moment I can't think of anything that could replace it, with a shorter word.
Do you make the correction, or do I have to do the Pull again??

Tensuko:
Better you do this

Roby1164:
ok

PR #2899, opened 2023-10-25
Driving beside combine improvements

schwiti6190:
- Changed the auto aim nodes to the fill root nodes
- Makes sure all fill units are checked including possible fill units of a truck

### DIFF
--- a/scripts/ai/AIDriveStrategyUnloadCombine.lua
+++ b/scripts/ai/AIDriveStrategyUnloadCombine.lua
@@ -300,7 +300,7 @@ function AIDriveStrategyUnloadCombine:setAIVehicle(vehicle, jobParameters)
     vehicle:resetCpCourses()
     self:resetPathfinder()
     --- Target nodes for unloading into the trailer.
-    self.trailerNodes = self:getTrailersTargetNodes()
+    self.trailerNodes = SelfUnloadHelper:getTrailersTargetNodes(vehicle)
 
 end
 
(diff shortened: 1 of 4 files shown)


scripts/ai/AIDriveStrategyUnloadCombine.lua
```diff
@@ -619,24 +652,37 @@ function AIDriveStrategyUnloadCombine:getCourseToAlignTo(vehicle, offset)
...
+    local trailers = AIUtil.getAllChildVehiclesWithSpecialization(self.vehicle, Trailer)
+    if not trailers then 
+        self:debugSparse('Can\'t find any trailers')
+        return
+    end
+    for _, trailer in pairs(trailers) do 
+        for ix, _ in pairs(trailer:getFillUnits()) do 
+            local node = trailer:getFillUnitExactFillRootNode(ix)
+            local _, _, trailerOffset = localToLocal(node, trailer.rootNode, 0, 0, 0)
+            local _, _, vehicleOffset = localToLocal(trailer.rootNode, self.vehicle.rootNode, 0, 0, 0)
```
pvaiko, line 668: If there is no guarantee that the tractor is not aligned with the trailer then I'd probably flip the nodes here to get the offset in the trailer's coordinate system for the case when the angle between the tractor and trailer is over 90 degrees.

scripts/ai/AIDriveStrategyUnloadCombine.lua
```diff
@@ -619,24 +652,37 @@ function AIDriveStrategyUnloadCombine:getCourseToAlignTo(vehicle, offset)
...
-        else
-            self:debugSparse('Combine says it can\'t load trailer')
-            --TODO: maybe then send the unloader away if activated?
         end
-    else
-        self:debugSparse('Can\'t find trailer')
     end
-    return trailer.rootNode
+    table.sort(targetNodes, function (a, b)
+        --- Sorts these nodes to make sure the front most node is the first.
```
pvaiko, line 681: I remember another issue, just can't find it, with overloading, where this sorting could be useful, maybe consider moving this into the SelfUnloadHelper so combines can also use it?

scripts/ai/AIDriveStrategyUnloadCombine.lua
```diff
@@ -712,40 +758,6 @@ function AIDriveStrategyUnloadCombine:isMoveablePipeDisabled()
...
-end
-
-function AIDriveStrategyUnloadCombine:isAutoAimNodeFixed()
-    return self.autoAimNodeFixed
-end
-
--- Make sure the autoAimTargetNode is not moving with the fill level (which adds realism trying to
--- distribute the load more evenly in the trailer but makes life difficult for us)
--- TODO: instead of turning it off completely, could try to reduce the range it is adjusted
-function AIDriveStrategyUnloadCombine:updateFillUnitAutoAimTarget(superFunc, fillUnit)
```
pvaiko, line 735: I was thinking of moving this to a specialization and limit the movement but what you did is a lot simpler.

Tensuko (approved):

pvaiko (approved):

Issue #2889, opened 2023-10-23
[BUG_SP] symmetric lane change setting is not working


WeAreV3N0M:
**READ FIRST**
https://github.com/Courseplay/Courseplay_FS22/wiki

**Steps to reproduce**
Unknown reason behind bug so not sure how to reproduce

**Actual behavior**
I have accepted a cotton harvesting job with two Case IH Module Express 635 harvesters (base game) And loaded a previously generated course that was generated for two machines. After the harvesters reached the end of first row (not counting headlands) both machines changed to the other side of the path for the next row. They turned from left side to right in next row and from right side to left. I tested it with symmetrical lane change turned on and turned off, it seemed to have no effect at all.

**Expected behavior**
With symmetrical lane change turned on, the machines should remain on the same side of the path all the way to the end. If started on the left side, after turning the worker should continue on left side.

**Courseplay version**
Version used:  7.3.1.3

**Log**
Please attach a log with all debug channels active.
For help with the debug channels,
please check out the readme at the github front page.

**Map**
Elm Creek Edit 4x V1.0.0.9 By Stevie

**Vehicles/ implements involved**
Case IH Module Express 635
[log.txt](https://github.com/Courseplay/Courseplay_FS22/files/13073636/log.txt)

Mod yes/no ?
no

**Relevant script mods**
Mods like Autodrive or VCA and so on.

RedMonster-HUN:
Just tested it with the John Deere machine. There is no difference between symmetrical lane change turned on or turned off.
[log.txt](https://github.com/Courseplay/Courseplay_FS22/files/13074055/log.txt)

Tensuko:
It is generally broken somehow.

rohne83:
```
Expected behavior
With symmetrical lane change turned on, the machines should remain on the same side of the path all the way to the end. If started on the left side, after turning the worker should continue on left side.
```

Nope, that's the behaviour for default asymmetric lane change, to be used for lands or spirals mode. With symmetric lane change (rather for updown or racetracks) the offset should of course always change at every turn. In my eyes that works as desired.

JDTiger:
Important addendum I found out:

When switching the lane change and/or Implement raise/lower you have to reload the course for the change to apply

WeAreV3N0M:
If I set one machine to go on left side and the other machine to the right, and then both try to go on right side (after me triple checking their CP settings are identical) that's everything but desired.
[...]

PR #2904, opened 2023-10-27
fix: symmetric lane change

pvaiko:
Only reuse a multitool offset course if all
parameters match.

#2889

### DIFF
--- a/scripts/Course.lua
+++ b/scripts/Course.lua
@@ -180,6 +180,15 @@ function Course:getMultiTools()
 	return self.multiTools or 1
 end
 
+---
+---@param multiTools number of tools
+---@param multiToolsPosition number which tool is this, same as position in calculateOffsetCourse()
+---@param multiToolsSameTurnWidth boolean turns are of the same width, ame as position in calculateOffsetCourse()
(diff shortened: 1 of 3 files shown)


schwiti6190 (approved):
Looks good just wondering if it is necessary to save and load multiToolsPosition and multiToolsSameTurnWidth to a xml file or not?

pvaiko:
> Looks good just wondering if it is necessary to save and load multiToolsPosition and multiToolsSameTurnWidth to a xml file or not?

I was wondering about that too, but the way it currently works is that we never save an offset course, we assume all saved courses are without offset. Also position/symmetric lane change isn't set on the course generator, they are on the HUD/vehicle settings, so what would you do with those when reloading an offset course?

Tensuko (approved):

schwiti6190:
> > Looks good just wondering if it is necessary to save and load multiToolsPosition and multiToolsSameTurnWidth to a xml file or not?
> 
> I was wondering about that too, but the way it currently works is that we never save an offset course, we assume all saved courses are without offset. Also position/symmetric lane change isn't set on the course generator, they are on the HUD/vehicle settings, so what would you do with those when reloading an offset course?

This makes sense and we don't plan to change it than this is fine :)

PR #2905, opened 2023-10-27
Separated the X tool offset for combine unloader and renamed Z tool offset(only for combine unloader)

schwiti6190:
- Saves the combine unloader x offset separate from any X tool offset
- Offset gets no longer reset between implement changes
- Might be a good idea to change the translations of combineOffsetX and combineOffsetZ as the current tooltips dosen't describe the functions of both.

### DIFF
--- a/config/MasterTranslations.xml
+++ b/config/MasterTranslations.xml
@@ -625,11 +625,11 @@
 			<Text language="de"><![CDATA[Versatz für den Ballensammler]]></Text>
 			<Text language="en"><![CDATA[Offset for the bale collector]]></Text>
 		</Translation>
-		<Translation name="CP_vehicle_setting_toolOffsetZ_title">
+		<Translation name="CP_vehicle_setting_combineOffsetZ_title">
 			<Text language="de"><![CDATA[Geräteversatz vertikal]]></Text>
 			<Text language="en"><![CDATA[Tool offset vertical]]></Text>
 		</Translation>
-		<Translation name="CP_vehicle_setting_toolOffsetZ_tooltip">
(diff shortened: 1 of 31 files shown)


config/MasterTranslations.xml
```diff
@@ -625,11 +625,11 @@
 			<Text language="de"><![CDATA[Versatz für den Ballensammler]]></Text>
 			<Text language="en"><![CDATA[Offset for the bale collector]]></Text>
 		</Translation>
-		<Translation name="CP_vehicle_setting_toolOffsetZ_title">
+		<Translation name="CP_vehicle_setting_combineOffsetZ_title">
 			<Text language="de"><![CDATA[Geräteversatz vertikal]]></Text>
 			<Text language="en"><![CDATA[Tool offset vertical]]></Text>
```
pvaiko, line 630: Unloader offset forward/backward?

config/MasterTranslations.xml
```diff
@@ -625,11 +625,11 @@
...
 		</Translation>
-		<Translation name="CP_vehicle_setting_toolOffsetZ_tooltip">
+		<Translation name="CP_vehicle_setting_combineOffsetZ_tooltip">
 			<Text language="de"><![CDATA[Geräteversatz vertikal]]></Text>
 			<Text language="en"><![CDATA[Tool offset vertical]]></Text>
```
pvaiko, line 634: Offset for the unloader when positioning under the pipe, forward +, backward - ?

pvaiko (approved):

config/MasterTranslations.xml
Tensuko: Would just write "Offset forward/backward" to save some space.
Same for the other, "Offset left/right".

PR #2975, opened 2023-11-23
Pathfinder refactoring

pvaiko:
I accidentally created a new branch for 2816 (lowercase i), instead of working on Issue-2816, but this gives us an opportunity to review the Pathfinder refactoring work I'm doing.

- [x] PathfinderContext moved out of PathfinderController
- [x] Remove PathfinderUtil.Context and use PathfinderContext instead
- [x] Remove the vehicle from the pathfinder calls (as it is part of the context)
- [ ] Refactor all pathfinder calls to use the PathfinderController (this may not all be done in this PR)

### DIFF
--- a/scripts/ai/AIDriveStrategyBunkerSilo.lua
+++ b/scripts/ai/AIDriveStrategyBunkerSilo.lua
@@ -579,8 +579,10 @@ function AIDriveStrategyBunkerSilo:startCourseWithPathfinding(course, ix, isReve
             setRotation(self.pathfinderNode.node, 0, yRot + math.pi, 0)
         end
 
+        local context = PathfinderContext(self.vehicle):allowReverse(true):offFieldPenalty(0) context:maxFruitPercent(math.huge)
+
         self.pathfinder, done, path = PathfinderUtil.startPathfindingFromVehicleToNode(self.vehicle, self.pathfinderNode.node,
-            0, 0, true)
+            0, 0, context)
 
         if done then
             return self:onPathfindingDoneToCourseStart(path)
@@ -629,8 +631,9 @@ function AIDriveStrategyBunkerSilo:startDrivingToParkPositionWithPathfinding()
     
         self:debug('Pathfinding to park position, with zOffset min(%.1f, %.1f)', -self.frontMarkerDistance, -steeringLength)
 
+        local context = PathfinderContext(self.vehicle):allowReverse(true):offFieldPenalty(0) context:maxFruitPercent(math.huge)
         self.pathfinder, done, path = PathfinderUtil.startPathfindingFromVehicleToNode(self.vehicle, self.parkNode,
-            0, 0, true)
+            0, 0, context)
 
         if done then
             return self:onPathfindingDoneToParkPosition(path)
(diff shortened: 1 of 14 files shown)


scripts/ai/AIDriveStrategyBunkerSilo.lua
```diff
@@ -579,8 +579,10 @@ function AIDriveStrategyBunkerSilo:startCourseWithPathfinding(course, ix, isReve
             setRotation(self.pathfinderNode.node, 0, yRot + math.pi, 0)
         end
 
-        self.pathfinder, done, path = PathfinderUtil.startPathfindingFromVehicleToNode(self.vehicle, self.pathfinderNode.node,
-            0, 0, true)
+        local context = PathfinderContext(self.vehicle):allowReverse(true):offFieldPenalty(0) context:maxFruitPercent(math.huge)
```
schwiti6190, line 582: ``context:maxFruitPercent(math.huge)`` Copy/paste error or intentionally?

scripts/ai/AIDriveStrategyBunkerSilo.lua
```diff
@@ -629,8 +631,9 @@ function AIDriveStrategyBunkerSilo:startDrivingToParkPositionWithPathfinding()
     
         self:debug('Pathfinding to park position, with zOffset min(%.1f, %.1f)', -self.frontMarkerDistance, -steeringLength)
 
-        self.pathfinder, done, path = PathfinderUtil.startPathfindingFromVehicleToNode(self.vehicle, self.parkNode,
-            0, 0, true)
+        local context = PathfinderContext(self.vehicle):allowReverse(true):offFieldPenalty(0) context:maxFruitPercent(math.huge)
```
schwiti6190, line 634: The same as above

scripts/ai/AIDriveStrategyBunkerSilo.lua
pvaiko: Good catch. What should it be? In startPathfindingFromVehicleToNode() it used to be `maxFruitPercent or (settings.avoidFruit:getValue() and 50 or math.huge)` ?

pvaiko: (So here, the equivalent would be `settings.avoidFruit:getValue() and 50 or math.huge`)

scripts/ai/AIDriveStrategyBunkerSilo.lua
pvaiko: Reply/question same as above.

scripts/ai/AIDriveStrategyBunkerSilo.lua
schwiti6190: Maybe we could add another simplified true or false behaviour for the fruit avoidence.
Then we could pass the setting value into it or use a strategy function to return true or false to the `context:avoidFruit()`.
`lua
PathfinderContext.DEFAULT_FRUIT_AVOIDENCE_PERCENTAGE = 50
--- Disables the fruit avoidance
function PathfinderContext:avoidFruit(avoidFruit[true/false])
    if avoidFruit then 
          self._maxFruitPercent = self.DEFAULT_FRUIT_AVOIDENCE_PERCENTAGE 
    else
          self._maxFruitPercent = math.huge
    end
    return self
end
`

pvaiko: We already have an ignoreFruit()

schwiti6190 (approved):

PR #2977, opened 2023-11-24
Added a stricter check for driving into the silo without pathfinding.

### LINKED ISSUE #2939: [BUG_SP] DLC Volvo wheel loader not path finding in silo load mode
**READ FIRST**
https://github.com/Courseplay/Courseplay_FS22/wiki

**Steps to reproduce**
Steps to reproduce the behavior.

Have Volvo LM 845 with volvo universial bucket lm841 do silo load mode

**Actual behavior**
A clear and concise description of what happen.

The problem is it keeps stopping "No Path" and there is a clear path for it i have turned off avoid driving in fruit 

**Expected behavior**

pathfind to the trailer

**Screenshots**
Screenshot (32)

**Courseplay version**
Version used: 7.3.1.3

**Log**
Please attach a log with all debug channels active.

[log.txt](https://github.com/Courseplay/Courseplay_FS22/files/13304545/log.txt)

For help with the debug channels,
please check out the readme at the github front page.

**Map**
Map played on. Alma Missouri

**Vehicles/ implements involved**
Which vehicle or implement?
Mod yes/no ?

DLC: Volvo LM845 With the Volvo Universal Bucket LM841

**Relevant script mods**
Mods like Autodrive or VCA and so on.

i have autodrive installed but was not using it for the unloader

schwiti6190:
Fixes a bug where the driver can't begin it's work correctly.
This could happen if the driver is started really close to the silo.

### DIFF
--- a/Courseplay.lua
+++ b/Courseplay.lua
@@ -195,6 +195,7 @@ function Courseplay:update(dt)
     g_devHelper:update()
     g_bunkerSiloManager:update(dt)
     g_triggerManager:update(dt)
+	g_baleToCollectManager:update(dt)
     if not self.postInit then 
         -- Doubles the map zoom for 4x Maps. Mainly to make it easier to set targets for unload triggers.
         self.postInit = true
@@ -214,6 +215,7 @@ function Courseplay:draw()
 		g_vineScanner:draw()
 		g_bunkerSiloManager:draw()
 		g_triggerManager:draw()
+		g_baleToCollectManager:draw()
 	end
 	g_devHelper:draw()
 	CpDebug:draw()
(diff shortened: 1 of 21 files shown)


pvaiko (commented):
Excuse my ignorance, I did not look into this problem in detail, but the log seems to complain about two issues: fruit at the target and collision with terrain and filltype: Corn. Are we sure this is not something the pathfinder logic could handle, like the areaToIgnoreFruit?

schwiti6190:
> Excuse my ignorance, I did not look into this problem in detail, but the log seems to complain about two issues: fruit at the target and collision with terrain and filltype: Corn. Are we sure this is not something the pathfinder logic could handle, like the areaToIgnoreFruit?

Oh you are right, I thought the problem was this part: ``vehicle off track, shutting off Courseplay now``.
I guess I didn't read the issue correctly...
This fix would still be a quality of life improvement, as the starting of the driver on an edge case angle relative to the silo might break the current logic on main.
For the issue I will wait for you're pathfinder adjustments then.

PR #2980, opened 2023-11-25
Fixes loaded settings values, like the x tool offset.

schwiti6190:
#1548 
Vehicle settings, that are changed based on attached implements, might change their loaded values automatically after loading. So we reset these settings on the first update tick back to the correct saved values.

### DIFF
--- a/scripts/specializations/CpCourseGeneratorSettings.lua
+++ b/scripts/specializations/CpCourseGeneratorSettings.lua
@@ -46,6 +46,7 @@ end
 
 function CpCourseGeneratorSettings.registerEventListeners(vehicleType)	
 	SpecializationUtil.registerEventListener(vehicleType, "onLoad", CpCourseGeneratorSettings)
+    SpecializationUtil.registerEventListener(vehicleType, "onUpdate", CpCourseGeneratorSettings)
     SpecializationUtil.registerEventListener(vehicleType, "onLoadFinished",CpCourseGeneratorSettings)
     SpecializationUtil.registerEventListener(vehicleType, "onCpUnitChanged", CpCourseGeneratorSettings)
 end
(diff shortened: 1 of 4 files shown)


scripts/specializations/CpCourseGeneratorSettings.lua
```diff
@@ -100,11 +101,16 @@ function CpCourseGeneratorSettings:onLoad(savegame)
...
-    if not spec.wasLoaded then
-        CpCourseGeneratorSettings.setAutomaticWorkWidthAndOffset(self)
+--- Apply auto work width after everything is loaded.
+function CpCourseGeneratorSettings:onLoadFinished()
+    CpCourseGeneratorSettings.setAutomaticWorkWidthAndOffset(self)
+end
+
+--- Resets the work width to a saved value after all implements are loaded and attached.
+function CpCourseGeneratorSettings:onUpdate(savegame)
+    if not self.finishedFirstUpdate then
```
pvaiko, line 111: When is this set?

scripts/specializations/CpVehicleSettings.lua
```diff
@@ -92,12 +92,21 @@ function CpVehicleSettings:onLoad(savegame)
     CpVehicleSettings.loadSettings(self, savegame)
 end
 
-function CpVehicleSettings:onLoadFinished()
-    local spec = self.spec_cpVehicleSettings
-    spec.wasLoaded = nil
+--- Resets the tool offset to a saved value after all implements are loaded and attached.
+function CpVehicleSettings:onUpdate()
+    if not self.finishedFirstUpdate then
```
pvaiko, line 97: When is this set?

Tensuko (approved):

PR #2990, opened 2023-12-03
Ai improvements 2023 12

pvaiko:
fix: harvester pocket creation

Back up further with longer vehicles before starting
the pocket, so they have room to drive fully into the
pocket.

fix: combine unloader stuck behind combine

When checking if we are in a good position behind
the combine to start unloading, allow for more
sideways error proportional to the distance from
the pipe, assuming we'll line up better as we
get closer.

fix: combine unloader moving out of combine's way

Fixed head-on and in-front/same direction check so
unloader will now reverse when behind the combine
(but has the same heading) to make way.

fix: towed harvester don't use combine headland turns

### DIFF
--- a/scripts/ai/AIDriveStrategyCombineCourse.lua
+++ b/scripts/ai/AIDriveStrategyCombineCourse.lua
@@ -146,7 +146,7 @@ function AIDriveStrategyCombineCourse:setAllStaticParameters()
     self.pullBackDistanceEnd = self.pullBackDistanceStart + 5
     -- when making a pocket, how far to back up before changing to forward
     -- for very long vehicles, like potato/sugar beet harvesters the 20 meters may not be enough
-    self.pocketReverseDistance = math.max(1.8 * AIUtil.getVehicleAndImplementsTotalLength(self.vehicle), 20)
+    self.pocketReverseDistance = math.max(1.9 * AIUtil.getVehicleAndImplementsTotalLength(self.vehicle), 20)
     -- register ourselves at our boss
     -- TODO_22 g_combineUnloadManager:addCombineToList(self.vehicle, self)
     self.waitingForUnloaderAtEndOfRow = CpTemporaryObject()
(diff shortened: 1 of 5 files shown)


scripts/ai/AIDriveStrategyUnloadCombine.lua
```diff
@@ -1583,29 +1588,33 @@ end
...
+        self:debug('%s is behind us, moving forward', CpUtil.getName(blockingVehicle))
         return Course.createFromNode(self.vehicle, self.vehicle:getAIDirectionNode(), 0,
                 frontMarkerOffset, frontMarkerOffset + self.maxDistanceWhenMovingOutOfWay, 5, false)
     end
 end
 
+------------------------------------------------------------------------------------------------------------------------
+-- Is the blocking vehicle in front of us?
+------------------------------------------------------------------------------------------------------------------------
+function AIDriveStrategyUnloadCombine:isBlockingVehicleAheadOfUs(blockingVehicle)
```
schwiti6190, line 1608: Maybe we could place this function in a util container, so we can reuse this in another strategy?

scripts/ai/AIDriveStrategyCombineCourse.lua
```diff
@@ -1331,7 +1332,10 @@ function AIDriveStrategyCombineCourse:startTurn(ix)
 
     -- Combines drive special headland corner maneuvers, except potato and sugarbeet harvesters
     if self.turnContext:isHeadlandCorner() then
-        if self.combineController:isPotatoOrSugarBeetHarvester() then
+        if self.combineController:isTowed() then
+            self:debug('Headland turn but this is a towed harvester using normal turn maneuvers.')
+            AIDriveStrategyCombineCourse.superClass().startTurn(self, ix)
+        elseif self.combineController:isPotatoOrSugarBeetHarvester() then
```
schwiti6190, line 1338: I guess we still need a replacement for `:isPotatoOrSugarBeetHarvester()`, as this function doesn't work for example for the same harvester, but as mods with different fill types.
Maybe a TODO comment, so we don't forget this?

schwiti6190 (changes requested):

Tensuko (approved):

PR #3022, opened 2023-12-19
Improving how root vegetable harvesters are recognized

tn4799:
Removed the old implementation with hardcoded fruit type checks. how searches all earth fruits and checks if at least one of them is supported. 

A earth fruit is defined by having a preparation entry in the xml file. The only exception to this definition is Sugarcane since it needs fruit preparing but is logially no earth fruit. This condition is checked in the seach.

Improvments of this PR:
- More Future proof and dynamically recognizion of earth fruits as long as giants sticks with fruit preparation
- now the self driving harvester of dewulf does turn properly

Things that still could be improved:
- Make chopper wait before turning to unload completly. This could be done to make turning a smooth movement without the unloader driving under the pipe all the time

### DIFF
--- a/scripts/CpUtil.lua
+++ b/scripts/CpUtil.lua
@@ -382,7 +382,7 @@ end
 --
 -- Preparation marks an earth fruit since this is the only factor that diversifies them from normal fruit types
 -- SUGARCANE is an exception and ignored
-function CpUtil.getAllEarthFruits()
+function CpUtil.getAllRootVegetables()
     local earthFruits = {}
     for _, fruitTypeData in pairs(g_fruitTypeManager.fruitTypes) do
         local minPreparingGrowthState = fruitTypeData.minPreparingGrowthState
--- a/scripts/CpUtil.lua
+++ b/scripts/CpUtil.lua
@@ -391,8 +391,9 @@ function CpUtil.getAllEarthFruits()
 
         -- check if fruit is needs herb removement to be harvested
         if minPreparingGrowthState ~= -1 and preparedGrowthState ~= -1 and name ~= "SUGARCANE" then
-            if g_fruitTypeManager:getFruitTypeByName(name) ~= nil then
-                table.insert(earthFruits, name)
+			local fruitType = g_fruitTypeManager:getFruitTypeByName(name)
+            if fruitType ~= nil then
+				table.insert(earthFruits, fruitType.index)
             end
         end
     end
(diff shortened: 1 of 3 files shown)


scripts/ai/controllers/CombineController.lua
```diff
@@ -128,12 +128,15 @@ function CombineController:isDroppingStrawSwath()
...
-function CombineController:isPotatoOrSugarBeetHarvester()
-    for i, fillUnit in ipairs(self.implement:getFillUnits()) do
-        if self.implement:getFillUnitSupportsFillType(i, FillType.POTATO) or
-                self.implement:getFillUnitSupportsFillType(i, FillType.SUGARBEET) then
-            self:debug('This is a potato or sugar beet harvester.')
-            return true
+function CombineController:isEarthFruitHarvester()
+    for _, fruitTypeIndex in pairs(CpUtil.getAllEarthFruits()) do
+        local fillUnitIndex = g_fruitTypeManager:getFillTypeIndexByFruitTypeIndex(fruitTypeIndex)
+        self:debug(string.format("check if fruitType %s is supported", g_fillTypeManager:getFillTypeNameByIndex(fillUnitIndex)))
```
pvaiko, line 134: No need for the string.format(), debug() accepts format strings and arguments.

scripts/CpUtil.lua
```diff
@@ -376,4 +376,27 @@ function CpUtil.isStateOneOf(myState, states)
         end
     end
     return false
+end
+
+-- Return all fruitTypeIndexes where a fruitType needs preparation. 
+--
+-- Preparation marks an earth fruit since this is the only factor that diversifies them from normal fruit types
+-- SUGARCANE is an exception and ignored
+function CpUtil.getAllEarthFruits()
```
pvaiko, line 385: Can we call it root vegetable instead of earth fruit? Earth fruit is not used in English.

pvaiko (changes requested):
I think this is a good temporary improvement.

On the long term, as we discussed with @schwiti6190 , it is not really the fruit being harvested that determines the type of turn, it is rather the vehicle properties, like length, turning radius, working width etc., which makes the crop harvester type turns useless.

scripts/CpUtil.lua
pvaiko: Also replacing all other occurrences of earth fruit?

schwiti6190 (approved):

PR #3124, opened 2024-03-09
Combine pre call option

MatteoDelfavero:
A new vehicle setting option is the combine’s filling level, which determines when to call the unloading car. Other language files are missing, Hungarian is ready

### DIFF
--- a/config/MasterTranslations.xml
+++ b/config/MasterTranslations.xml
@@ -831,11 +831,11 @@
 		</Translation>
 			<Translation name="CP_vehicle_setting_callUnloaderPercent_title">
 			<Text language="de"><![CDATA[Brief vor dem Anruf]]></Text>
-			<Text language="en"><![CDATA[Pre call level]]></Text>
+			<Text language="en"><![CDATA[Unloader call level]]></Text>
 		</Translation>
 		<Translation name="CP_vehicle_setting_callUnloaderPercent_tooltip">
 			<Text language="de"><![CDATA[Loader-Aufruf oberhalb des eingestellten Levels.]]></Text>
-			<Text language="en"><![CDATA[Loader call above set level.]]></Text>
+			<Text language="en"><![CDATA[Call unloader when the harvester fill level is above this limit.]]></Text>
 		</Translation>
 		<Translation name="CP_vehicle_setting_strawSwath_onlyCenter">
 			<Text language="de"><![CDATA[Vorgewende deaktiviert]]></Text>
--- a/config/MasterTranslations.xml
+++ b/config/MasterTranslations.xml
@@ -829,6 +829,14 @@
 			<Text language="de"><![CDATA[Strohablage an, aus oder nur im Vorgewende aus.]]></Text>
 			<Text language="en"><![CDATA[Enables/disables straw swath or only disables it on headlands.]]></Text>
 		</Translation>
+			<Translation name="CP_vehicle_setting_callUnloaderPercent_title">
+			<Text language="de"><![CDATA[Brief vor dem Anruf]]></Text>
+			<Text language="en"><![CDATA[Pre call level]]></Text>
+		</Translation>
+		<Translation name="CP_vehicle_setting_callUnloaderPercent_tooltip">
+			<Text language="de"><![CDATA[Loader-Aufruf oberhalb des eingestellten Levels.]]></Text>
+			<Text language="en"><![CDATA[Loader call above set level.]]></Text>
+		</Translation>
 		<Translation name="CP_vehicle_setting_strawSwath_onlyCenter">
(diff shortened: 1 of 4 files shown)


Tensuko:
That won't work.
Our complete system is set for the arround 80% pre calculation.
We did not add that setting on purpose!

config/MasterTranslations.xml
```diff
@@ -828,6 +828,14 @@
 		<Translation name="CP_vehicle_setting_strawSwath_tooltip">
 			<Text language="de"><![CDATA[Strohablage an, aus oder nur im Vorgewende aus.]]></Text>
 			<Text language="en"><![CDATA[Enables/disables straw swath or only disables it on headlands.]]></Text>
+		</Translation>
+			<Translation name="CP_vehicle_setting_callUnloaderPercent_title">
+			<Text language="de"><![CDATA[Brief vor dem Anruf]]></Text>
+			<Text language="en"><![CDATA[Pre call level]]></Text>
```
pvaiko, line 834: Unloader call level

config/MasterTranslations.xml
```diff
@@ -828,6 +828,14 @@
...
+		</Translation>
+		<Translation name="CP_vehicle_setting_callUnloaderPercent_tooltip">
+			<Text language="de"><![CDATA[Loader-Aufruf oberhalb des eingestellten Levels.]]></Text>
+			<Text language="en"><![CDATA[Loader call above set level.]]></Text>
```
pvaiko, line 838: Call unloader when the harvester fill level is above this limit.

scripts/ai/AIDriveStrategyCombineCourse.lua
```diff
@@ -103,7 +103,6 @@ function AIDriveStrategyCombineCourse:init(task, job)
         }
     })
 
-    -- TODO: move this to a setting
     self.callUnloaderAtFillLevelPercentage = 80
```
pvaiko, line 106: We can remove this variable and use the setting everywhere instead.

scripts/ai/AIDriveStrategyCombineCourse.lua
```diff
@@ -722,6 +721,7 @@ function AIDriveStrategyCombineCourse:estimateDistanceUntilFull(ix)
     end
     local litersUntilFull = capacity - fillLevel
     local dUntilFull = litersUntilFull / self.litersPerMeter
+    self.callUnloaderAtFillLevelPercentage = self.settings.callUnloaderPercent:getValue()
```
pvaiko, line 724: use directly the setting here, and the debug() below

config/VehicleSettingsSetup.xml
```diff
@@ -95,6 +95,8 @@
 		<Setting classType="AIParameterBooleanSetting" name="selfUnload"/>
 		<!--Unload on first Headland-->
 		<Setting classType="AIParameterBooleanSetting" name="unloadOnFirstHeadland" defaultBool="false" isExpertModeOnly="true"/>
+		<!--UnloadLevel-->
+		<Setting classType="AIParameterSettingList" name="callUnloaderPercent" min="0" max="100" incremental="5" default ="80" unit="4" isExpertModeOnly="false"/>
```
pvaiko, line 99: min should not be 0 (no idea if the code can handle that case), should be some reasonable value, like 50?

pvaiko (changes requested):
Some changes needed, otherwise fine. I leave the German translation to @Tensuko

scripts/ai/AIDriveStrategyCombineCourse.lua
MatteoDelfavero: I looked after pull request to see if this is really unnecessary, you're right.

config/VehicleSettingsSetup.xml
MatteoDelfavero: Tested and it works perfectly, just like in Autodrive.

MatteoDelfavero:
> That won't work. Our complete system is set for the arround 80% pre calculation. We did not add that setting on purpose! It could only be set save between 70% and 80% and this is not a big difference so we did not add this at all.

I have been using this modification on my server for 2 days (harvesting about 20+ lands) and it works perfectly.

Tensuko:
> > That won't work. Our complete system is set for the arround 80% pre calculation. We did not add that setting on purpose! It could only be set save between 70% and 80% and this is not a big difference so we did not add this at all.
> 
> I have been using this modification on my server for 2 days (harvesting about 20+ lands) and it works perfectly.

What setting? 10%? 20% ? 100%?
Won't go higher then 90% and not lower then 60% change that min and max to those values and I give it a try. :)

Issue #3181, opened 2024-03-30
[BUG_SP] Invalid Crop Type Error Message


lcc1211:
**READ FIRST**
https://github.com/Courseplay/Courseplay_FS22/wiki

**Steps to reproduce**
Set up CP to offload a combine. Using Giants offloader, and send to a sell point.

**Actual behavior**
Once the trailer is full it drives to a sell point and unloads. Once the whole load is sold and the vehicle is ready to drive back to the field CP stops and an error message is displayed saying the crop type is invalid for the target. At first I thought this might be a Giants issue, but I created a non-cp job where the vehicle takes something the silo to the sell point. It does this without any problem.

This did happen one time when I first set up a course. When I set it to start CP stopped immediately and displayed the message. 

**Expected behavior**
Offload the crop at the sell point and return to the field.

**Screenshots**
I tried getting a screenshot several times, but the error message doesn't show.

**Courseplay version**
Version used: 7.4.1.1

**Log**
Please attach a log with all debug channels active.
For help with the debug channels,
please check out the readme at the github front page.

**Map**
Iowa Plains

**Vehicles/ implements involved**
Which vehicle or implement?
Mod yes/no ? Yes, truck is on Modhub - Case IH S1800+. The trailer is a mod as well - Demmler Der-Siler-Profi. Although this is the same equipment I was using with earlier versions of CP. 

**Relevant script mods**
Mods like Autodrive or VCA and so on. = Neither of these or any other "driving" mods.
[log.txt](https://github.com/Courseplay/Courseplay_FS22/files/14812100/log.txt)

lcc1211:
I did a bit of troubleshooting this morning. I removed the mod sugar beet and potato harvesters that allowed them to harvest parsnips, carrots etc., and the straw mod hoping it was one of those causing the issue, but the error message was still there. I rolled my CP version back to 7.4.1.0 and the error message is no longer there. The vehicle leaves the sell point and returns to the field waiting to offload again.

PR #3192, opened 2024-04-06
Only check for valid filltype while driving to the unloading point

schwiti6190:
#3181

@lcc1211 Can you checkout this fix ?


lcc1211:
Looking good now. No issues when dropping at a sell point on any other point accepting the crop.

PR #3376, opened 2024-09-01
Update translation_it.xml

Roby1164:


Tensuko:
Thank you.

Roby1164:
It would be nice to have advance notice so that the translation can be included in the release.

Tensuko:
To be honest, not everyone is as fast as you and we don't want to wait for all translater to provide bugfixes and stuff.
The Version won't get send to ModHub yet, we ofc wait for translations and we might need to fix some more bugs first so there is the timeslot for our translator.

PR #3531, opened 2024-10-14
Added missing Refill controllers and a few code improvements

schwiti6190:
Test:
- Sowing machine
- Sprayer
- Sowing machine with sprayer
- Forage wagon
- Chopper
- Baler
- Tree planter


scripts/ai/controllers/BalerController.lua
```diff
@@ -28,6 +28,10 @@ function BalerController:init(vehicle, baler)
     self.baleWrapperSpec = self.baler.spec_baleWrapper
     self.lastDroppedBale = CpTemporaryObject()
     self:debug('Baler controller initialized')
+    local additives = self.balerSpec.additives
```
pvaiko, line 31: Does it make sense to move additives to the ImplementController so we don't need this boilerplate code everywhere?

PR #3584, opened 2024-11-04
Pocket

pvaiko:


scripts/util/HelperNode.lua
```diff
@@ -0,0 +1,110 @@
...
+    self:place(x, y or 0, z)
+end
+
+--- Get the position of a point relative to node, in the helper node's coordinate system.
+---@param node number
+---@param lx number x coordinate of the point relative to the node
+---@param ly number y coordinate of the point relative to the node
+---@param lz number z coordinate of the point relative to the node
+---@return number, number, number
+function HelperTerrainNode:localToLocal(node, lx, ly, lz)
```
schwiti6190, line 108: This function is only working for HelperTerrainNode and not for the base HelperNode

scripts/util/HelperNode.lua
```diff
@@ -0,0 +1,110 @@
...
+---@class HelperNode
+HelperNode = CpObject()
+
+--- Create a new helper node, linking it to the given rootNode.
+---@param name string
+---@param rootNode number|nil
+function HelperNode:init(name, rootNode)
+    self.node = createTransformGroup(name)
+    self.rootNode = rootNode
+    link(self.rootNode, self.node)
```
schwiti6190, line 33: What happens here if rootNode is nil?

scripts/util/HelperNode.lua
```diff
@@ -0,0 +1,110 @@
...
+function HelperNode:draw(text)
+    if entityExists(self.node) then
+        DebugUtil.drawDebugNode(self.node, text or getName(self.node), false, 0)
+    end
+end
+
+-- A helper node that is linked to the terrain root node, and the y coordinate is always
+-- the terrain height at the x, z position.
+---@class HelperTerrainNode : HelperNode
+HelperTerrainNode = CpObject(HelperNode)
```
schwiti6190, line 66: Not sure if inheritence is the best approach here.
I would prefer something like this: 
```lua
-- Creates a new helper node linked to the terrain root node.
function HelperNode.createTerrainNode(name )
  return HelperNode(name, g_currentMission.terrainRootNode)
end
```

scripts/util/HelperNode.lua
```diff
@@ -0,0 +1,110 @@
...
+function HelperTerrainNode:init(name)
+    HelperNode.init(self, name, g_currentMission.terrainRootNode)
+end
+
+--- Place the node at the given position and rotation.
+---@param x number
+---@param y number if the node is linked to the terrain, this is relative to the terrain height at x, z
+---@param z number
+---@param yRotation number|nil Rotation set only if not nil
+function HelperTerrainNode:place(x, y, z, yRotation)
```
schwiti6190, line 79: I would prefer this functionality in a separate function like so maybe:
```lua
---@param yRotation number|nil
---@param yOffset number|nil
function HelperNode:placeTerrain(x, z, yRotation, yOffset)

end
```

scripts/util/HelperNode.lua
```diff
@@ -0,0 +1,110 @@
...
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+]]--
+
+-- A helper node is a temporary node to help with relative vehicle position calculations.
+-- It wraps the creation of the Giant's engine node.
+-- NOTE: it must be destroyed explicitly to avoid node leak.
+
+---@class HelperNode
```
schwiti6190, line 24: This is a fantastic idea :)
Maybe you got the motivation to add missing commanly used functions, like worldToLocal(), localDirectionToWorld().

scripts/util/HelperNode.lua
pvaiko: Was struggling with this as well, probably we don't even need a generic HelperNode (at least as of now, I'm not aware of a use case), and should only implement a HelperTerrainNode.

scripts/util/HelperNode.lua
pvaiko: I had something similar first, the reason I don't like it, is that the behavior/interpretation for the y coordinate is different.

schwiti6190 (approved):

PR #3591, opened 2024-11-06
fix typo

dobasy:
I think this is typo. Sorry if it's not. Feel free to close this if you want to make your own commit :)


Tensuko:
it is indeed, but that PR won't change the text, I will fix it for you. Thanks anyway.

dobasy:
Thanks! But I think there's new problem for non-English languages...
スクリーンショット (525)

PR #3597, opened 2024-11-13
Fs25

pvaiko:
Just a quick pass to make the game run up with Courseplay. Still throwing many errors, mainly GUI. 

The good people at Giants do not bother to publish documentation, scripts or debugger, as usual, why would they...


Yboux:
So many people using CoursePlay, they could make an effort at least.

PR #3620, opened 2024-12-23
Update translation_nl.xml

Petje81:
Translation update till row 344


Tensuko:
Wrong repository.
https://github.com/Courseplay/Courseplay_FS25

Petje81:
Might be something i did wrong but i forked it from the FS25 and it pushed it to FS22 in github, even though i assigned it (or thought i did) it still put it in FS22

